PR #5, opened 2016-04-05
Added sample output to README

ingve:
When evaluating various disassembly libraries, I wanted to see how the output from Zydis looked with the default formatter. If sample output from the example program is added to the README file, potential users won't have to download and build Zydis just to find out if the output is suitable for their needs.


athre0z:
Good point, thanks! We replaced the example code with something that has a few more example instructions to make it a little more interesting.

PR #10, opened 2016-12-07
Added #undef for IN and OUT in case they were defined by <minwindef.h>

reliasn:
If you
```
#include <Windows.h>
```
before
```
#include <Zydis.hpp>
```
the mnemonics IN and OUT conflict with the macros defined in <minwindef.h>:
```
#ifndef IN
#define IN
#endif

#ifndef OUT
#define OUT
#endif
```
If Zydis should by design be included first, then I guess we can ignore this PR.

### DIFF
--- a/Zydis/ZydisOpcodeTable.hpp
+++ b/Zydis/ZydisOpcodeTable.hpp
@@ -39,6 +39,14 @@
 #include <stdint.h>
 #include <cassert>
 
+#ifdef IN
+#undef IN
+#endif
+
+#ifdef OUT
+#undef OUT
+#endif
+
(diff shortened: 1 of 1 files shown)


athre0z:
We're aware of this collision on Windows and also considered undefing those, but decided it probably wouldn't be a good idea to do it in our code as it probably breaks a few Windows headers. Instead, users should either do it themselves when they include zydis or just alter the include order.

We're currently working hard on getting version 2, a fully rewritten version of the library, ready for release. In that new version, all of our symbols have a common prefix, avoiding such collisions all together.

Anyways, thanks for the contrib!

PR #18, opened 2017-10-14
Fixed an uninitialized variable in ZydisChangeCase

mrexodia:
This code didn't compile with MSVC2013 (AppVeyor integration perhaps to prevent this in the future?)


athre0z:
Phew, wtf. That didn't even print a warning with Clang. And it even worked. Thanks!

PR #19, opened 2017-10-14
Convert all functions in ZydisFormatter to take const arguments

mrexodia:
Relevant discussion in the x64dbg channel:

```
<mrexodia> is there a reason that the formatter doesnt take constant pointers everywhere?
<athre0z> there are userdata pointers in the instruction and operand structs, which may be read & written by user provided hooks
<athre0z> personally, I'd prefer them as part of the formatter or an extra argument to the FormatEx func instead of having non-const-ptrs everywhere
<mrexodia> @athre0z I don't think formatters would actually modify the pointers, probably just cast them to a struct/class and modify that (which is fine since the pointer in the struct can be void* which is non-const)
```

### DIFF
--- a/include/Zydis/Formatter.h
+++ b/include/Zydis/Formatter.h
@@ -326,7 +326,7 @@ enum ZydisDecoratorTypes
     ZYDIS_DECORATOR_TYPE_MAX_VALUE = ZYDIS_DECORATOR_TYPE_EVICTION_HINT
 };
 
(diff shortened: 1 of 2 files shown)


flobernd:
Thanks for the PR. I merged your changes and added `userData` from `ZydisDecodedInstruction` as an additional parameter to all of the formatter-callbacks.

mrexodia:
Great, thanks!

Issue #16, opened 2017-07-28
Errors while using Zydis.lib in windows driver


pakt:
Quick fixes for: 
* `stdint.h` not found -> just add a copy from visual studio
* `unresolved external symbol __imp_wassert`: redefine `ZYDIS_ASSERT` and `ZYDIS_UNREACHABLE` to `;`
* `unresolved external symbol __imp___stdio_common_vsprintf`: use `winkernel_mm.*` from [capstone](https://github.com/aquynh/capstone/tree/master/windows)
* missing `DriverEntry` / `ALIGN` errors -> see [this issue](https://github.com/aquynh/capstone/issues/965)
* `unresolved external symbol __imp___stdio_common_vsprintf`: add `/D _NO_CRT_STDIO_INLINE` to `cl` options in Zydis project

athre0z:
Thanks, I'll take a look!

pakt:
I'm using Visual Studio Community 2015, v14.

athre0z:
`stdint.h` and the assert should be easy fixes. Regarding `vsprintf`, we'll sooner or later just get rid of the dependencies on the CRT formatting functions. For now, I'll probably hack together something similar to what Capstone does. 

The lack of `DriverEntry`, I wouldn't call that a bug. I mean, compiling Zydis to a driver on its own doesn't make a lot of sense. `DriverEntry` should be defined by whatever driver is using Zydis by statically linking against it. Maybe I'm missing something, I don't have a lot of experience with Windows drivers.

pakt:
Regarding `DriverEntry`: I had a driver with custom entry point (different than `DriverEntry`). Citing my capstone issue:


> Using `cs_malloc` in your driver can trigger `(KernelBufferOverflowLib);` to be linked in. `(KernelBufferOverflowLib);` uses a wrapper for `DriverEntry`. If there is no `DriverEntry` defined in your code, linking fails. Solution is to use "DriverEntry" as your entrypoint (custom name will fail).

athre0z:
My WDK managed to corrupt itself during installation, I can't even build freshly created driver projects. Thanks MS. I'll remove all Windows SDKs and WDKs I have installed, apply fresh ones and try again. May take a while. Pretty sure I'll end up trashing my entire VM, setting up a new one.

pakt:
I followed "Build" instructions from [here](https://github.com/tandasat/HyperPlatform) and it worked.

athre0z:
[...]

PR #24, opened 2017-11-29
Visual Studio project files / Windows kernel mode sample

Mattiwatti:
(Also see #16 for some of the reasoning and some discussion re: this.)

This PR adds an MSVC solution and project files in '/msvc' under the root directory. While CMake is also able to generate MSVC files, this approach offers some benefits:
- The ability to combine 32 and 64 bit configurations in one project.
- The ability to combine static and dynamic build configurations in one project.
- Support for kernel mode toolsets (i.e. the WDK).
- More fine-grained control over project settings that affect compiler flags than what is offered by CMake.

The drawback of course is that this additional set of files needs to be maintained. I'd be OK with doing this. It'd be nice if there was some automated way for me to receive a ping if CMakeLists.txt gets updated so I can see if anything needs changing. But I'm not exactly an expert on CI setups, so hopefully someone has an idea how to set this up.

The solution contains static, dynamic and kernel build configurations. Some more details can be found in /msvc/README.txt. Building the kernel configuration requires the [Windows Driver Kit](https://go.microsoft.com/fwlink/p/?linkid=859232).

I've added a small kernel mode example project for a driver that disassembles its own entry function and prints the output. Feedback welcome :)


athre0z:
Hmm.. my VS2017 (on Win10 1709) refuses to open the properties for the Zydis project within the solution:
e0m7y9cisur1ihsr

For the first few times, it even crashed my whole VS ("has encountered an error and had to be closed").

Also, there is this message box when opening the solution.
9aemsrowc0ta1igc

Did you verify that is actually works without WDK? In case you did, it might be related to me having an older WDK installed that doesn't yet support VS2017. Installing the new WDK fixed the issue, by the way.

Mattiwatti:
Yes, this is a known (and very annoying) issue that has only been fixed a couple of weeks ago with the new RS3 WDK (10.0.16299.0). So if your WDK version is older than that, you'll have to remove it first and then install the new version (definitely in that order, trust me :P).

I did test the project on a VM without the WDK installed - it worked fine, sans being able to compile the kernel configuration of course.

athre0z:
Alright, nice. We might want to put a warning about this into the README, since those errors aren't especially self explanatory and I guess it might affect quite some people (given Zydis user base are usually low level peope).

Mattiwatti:
Good point, done.

athre0z:
Alright, last thing before we merge this: you're disabling stack cookies and CFG for release binaries. I really don't feel comfortable with that. In the original issue, you argued that static analysis would deprecated them, but according to our experience, this isn't (yet?) true. They do a pretty okayish job at spotting obvious buffer overflows, but they also miss stuff quite frequently. 

For example, [open up `Decoder.c`, switch to line 323](https://github.com/zyantific/zydis/blob/master/src/Decoder.c#L323) and mess up this if condition. Even delete it. MSVC, Clang and Coverity static analysis will stay silent, despite this causing an arbitrary sized stack overflow (overflowing `instruction->data` array) if the instruction structure is allocated on the stack (or a heap overflow if malloced).

ijy3ob0ij0g4z9rv

The program in the background is XCode, which uses Clang's static analysis. This was a actual bug in [an earlier alpha version](https://github.com/zyantific/zydis/commit/867b6bc1094c29a7adef993a8964cf2421a8cdaa#diff-a1dc2ccfa8a4f5a4bd9b00b63cfb6329), btw, overflowing to a maximum of 15 bytes.

Would you be okay with enabling them again?

Mattiwatti:
Yes, of course. Actually it was never my intention to disable these solution-wide; only for the ZydisWinKernel sample. I must have set these flags while having all of the configurations selected rather than just the driver.

The reasoning for disabling /GS and CFG in the driver is mostly for simplicity: different compilers implement stack cookies differently, and there is a linker component involved as well which has changed a few times with different DDK/WDK releases. Currently it looks like this:
```asm
; NTSTATUS __stdcall GsDriverEntry(PDRIVER_OBJECT DriverObject, PUNICODE_STRING RegistryPath);
mov [rsp+8], rbx
push rdi
sub rsp, 0x20
mov rbx, rdx
mov rdi, rcx
call __security_init_cookie ; <-- this does NOTHING since the kernel now initializes the cookie!
mov rdx, rbx
mov rcx, rdi
call DriverEntry
mov rbx, [rsp+0x30]
add rsp, 0x20
pop rdi
retn
```
But there have also been versions that use a `jmp` to call the driver entry point. Since these aren't very interesting functions to disassemble, you would then have to take into account the possible ways `GsDriverEntry` transfers execution to `DriverEntry` to get to the real entry point, and then preferably exclude it from the input buffer as well (note that `GsDriverEntry` can come either before or after `DriverEntry` in the PE section). This is of course all not overly difficult, but my goal was mostly to create a very simple 'how to build a driver with Zydis' sample that also prints some more interesting stuff as a side effect :) E.g.:

dbgview

The same reasoning applies to CFG, which transforms regular `call` instructions into mini-puzzles whose pieces are divided over different functions and even sections.

As for stack cookies in general... heh. I guess we favour different types of bugs :) I honestly can't remember the last time I encountered a stack buffer overrun (or at least not one that /GS caught, and I do have it enabled because my boss would probably fire me for even suggesting to disable them.) On the other hand, my test system bugchecks due to null pointer dereferences at least three times a day.

athre0z:
Ah, makes sense, yeah. Nevertheless, people tend to C&P sample projects to start off and I think it's better having all protections enabled. In your latest commit, CFG is still disabled in the kernel driver sample. Is this intentional for the more interesting output? A slightly different idea would be to simply do a `KernelGetProcAddress` on a kernel export of your choice for some more interesting output instead.

Mattiwatti:
Yes, this is intentional, but it does not actually affect the output of the binary because the default setting for CFG is already disabled (a good decision overall I think - CFG is a fairly mediocre mitigation policy for security vulnerabilities after-the-fact, and it does not help in finding bugs that also happen to be potential security vulnerabilities in the way that /GS does). I'm just somewhat wary of settings like these that I for sure do not ever want to have enabled in my programs, so I set them to disabled to head off any unwanted changes in the VS defaults in the future. But that's just my take on it. If you prefer it on the 'default-off' setting or even enabled, that's fine too.

I could use `MmGetSystemRoutineAddress` on a kernel function, yes, but I actually thought it was sort of neat to have a program that disassembles itself. Plus you can actually follow along with the source code, which you definitely can't do with raw disassembly of the Windows kernel that's also missing symbol names.

athre0z:
If it is default disabled anyways, I don't care.

Very well! Thank you for your contrib and your willingness to maintain this!

I noticed that you put all Zydis headers into all example and tool projects, probably in order to mirror CMake behavior. Thats more or less just a side affect fron how CMakes `target_link_library` works, not really a preference of ours. If it makes things easier for you, feel free to not include them whenever an update is required the next time. Whatever you prefer.

Mattiwatti:
Awesome!

Re: including all headers in all projects: I do have a slight preference for including external header dependencies in a project (meaning headers within the solution itself, I don't add Windows.h and stdio.h obviously), but for standalone projects I would usually put these under a separate filter in VS, e.g. `/include/Zydis`. Maybe this style would be better, to make it clear that these headers are part of the public Zydis library ones than headers belonging to the example project itself (at the moment `FormatHelper.h` is the only such file.)

PR #25, opened 2017-12-02
Add hook for operand seperator to formatter

th0rex:


athre0z:
Thanks! Merged (with rebase -i -- GH failed to detect it as merge).

PR #26, opened 2017-12-10
Update MSVC project files

Mattiwatti:
Hi,

This updates the MSVC project files to match the current CMakeLists.txt. I've removed the 'Header Files\Zydis' project filter and moved these to simply 'Header Files', since there are now no longer any headers under `/src`.

### DIFF
--- a/msvc/examples/FormatterHooks.vcxproj
+++ b/msvc/examples/FormatterHooks.vcxproj
@@ -400,7 +400,8 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="..\..\examples\FormatterHooks.c" />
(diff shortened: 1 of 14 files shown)


athre0z:
Merged, thanks! In case you were wondering why we didn't bump your PR yet, we thought we'd spare you the pain of adapting to all "in between versions" changes and just bump before we plan to release a new tag.

Mattiwatti:
Sure, that's cool. It doesn't really matter that much to me - if anything I prefer having to update slightly more often with incremental changes, than doing `git merge` and going "...oh. so that's what happened.". But I see I timed my PR right, as you just merged develop into master ;)

PR #35, opened 2018-02-16
Fix out-of-order entries in register map

### LINKED ISSUE #33: Incorrect register class of instruction pointer register
```
$ ZydisInfo -64 488b05b6310500
...
== [ OPERANDS ] ============================================================================================
##       TYPE  VISIBILITY  ACTION      ENCODING   SIZE  NELEM  ELEMSZ  ELEMTYPE                        VALUE
--  ---------  ----------  ------  ------------   ----  -----  ------  --------  ---------------------------
 0   REGISTER    EXPLICIT       W     MODRM_REG     64      1      64       INT                          rax
 1     MEMORY    EXPLICIT       R      MODRM_RM     64      1      64       INT  TYPE  =                 MEM
                                                                                 SEG   =                  ds
                                                                                 BASE  =                 rip
                                                                                 INDEX =                none
                                                                                 SCALE =                   0
                                                                                 DISP  =  0x00000000000531B6
```

Here this instruction's memory operand uses `%rip` as its base register. If I modify ZydisInfo to print the register class, as in `ZydisRegisterGetClass(instruction->operands[i].mem.base)`, I get `ZYDIS_REGCLASS_INVALID`, which seems incorrect. I expect `ZYDIS_REGCLASS_IP`.

rgov:
`ZydisRegisterGetClass()` will fail to map some registers to the correct class because the entries in `registerMap` are out of order. Fixes #33.


rgov:
Already fixed in 8c5e216e11fd478bc027dc33bffcaae788a0903f on `develop`. Closing.

PR #47, opened 2018-07-11
Some FreeBSD fixes 

gdbinit:
The ZydisPerfTest.c doesn't compile in FreeBSD because it lands into unsupported platform case.
And including Zydis static library in other shared or static libraries fails because missing -fPIC (not required in Linux or OS X).

Not sure if it fits your preferred style, feel free to modify :-)


athre0z:
Nope, looks good! Merged, thanks!

PR #48, opened 2018-07-13
Pe example fix

awant:
Previous version of pe disasm file was too old and didn't compiled with the new code of zydis.
Fix this with adding some useful functions for work with pe file.

### DIFF
--- a/examples/ZydisPE.c
+++ b/examples/ZydisPE.c
@@ -1,29 +1,30 @@
 /***************************************************************************************************
 
-  Zyan Disassembler Engine (Zydis)
+Zyan Disassembler Engine (Zydis)
 
-  Original Author : Florian Bernd
+Original Author : Florian Bernd
 
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
- * in the Software without restriction, including without limitation the rights
- * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


flobernd:
@athre0z is correct about the rewrite. I already updated the old `ZydisPE` example to be compatible with the future branch, but did not commit it yet. There are some things I have to change before pushing it - like getting rid of `Windows.h` to make the program platform independent (your RVA to file-offset  conversion might actually become handy, as I won't be able to use `LoadLibrary`).

awant:
I started to use Zydis in my own project and had to rewrite code with pe parsing. So it is pretty close as I use this. In my project I solved problem with Windows.h, but it is not very convenient.
I can try to help you with it, but It will be very helpful to see the current state of update.

flobernd:
Thanks for the kind words and your help offer! I pushed the `ZydisPE` tool to the future branch (c2836053c388d7e052c2f03455f894bf1613b677). It should compile and work fine as long as you don't load 32-bit files with the 64-bit version of the tool and vice versa.

flobernd:
I pushed my latest local changes. The tool should now be platform independent (in theory; only tested on Windows and Linux), but the code is not very nice yet (lots of duplicated lines due to different data-structures for 32/64-bit PE-files). Not sure if there is a more elegant way ...

Mapping the complete PE-file into memory using `fopen` and `malloc` might not be the smartest idea as well.

PR #66, opened 2018-10-30
Fix ZydisOperandAction enum

zlaski:
The ZydisOperandAction values must be encoded as bits to be usable.

This might be applicable to the new version of Zydis.

There are also changes to the ZydisDisasm and ZydisInfo project files so that include paths do not contain $(SolutionDir).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
 include(GenerateExportHeader)
 include(GNUInstallDirs)
 
-project(Zydis VERSION 2.0.2)
+project(Zydis VERSION 2.1.0)
 
 # =============================================================================================== #
 # Overridable options                                                                             #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
 include(GenerateExportHeader)
 include(GNUInstallDirs)
 
-project(Zydis VERSION 2.1.0)
+project(Zydis VERSION 2.0.2)
 
 # =============================================================================================== #
 # Overridable options                                                                             #
--- a/include/Zydis/Formatter.h
+++ b/include/Zydis/Formatter.h
@@ -632,8 +632,8 @@ typedef ZydisStatus (*ZydisFormatterDecoratorFunc)(const ZydisFormatter* formatt
 struct ZydisFormatter_
 {
     ZydisLetterCase letterCase;
-    ZydisBool forceMemorySegment;
-    ZydisBool forceMemorySize;
+    ZydisBool forceorySegment;
+    ZydisBool forceorySize;
     ZydisU8 formatAddress;
     ZydisU8 formatDisp;
     ZydisU8 formatImm;
--- a/include/Zydis/Formatter.h
+++ b/include/Zydis/Formatter.h
@@ -114,7 +114,7 @@ enum ZydisFormatterProperties
     /**
      * @brief   Controls the format of addresses.
      *
-     * The default value is `ZYDIS_ADDR_FORMAT_JMP_ABSOLUTE`.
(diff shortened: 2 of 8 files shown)


zlaski:
Thank you for the clarification regarding ZydisOperandAction in v2.  When I get a chance, I’ll switch to v3 and pull in your mods.

I’ve glanced at your header files and I noticed you’ve changed how enums are declared, and have a few suggestions.  In C++11, you can declare the underlying size of the enum, like so (to avoid bloating the ZydisDecodedInstruction object, which my project will have tons of):

typedef enum ZydisMachineMode_: ZydisU8
{
   :

For improved type safety (and _way_ easier debugging), you may do

typedef enum class ZydisMachineMode_: ZydisU8
{
   :

The foregoing will require you to explicitly qualify any enum use, e.g., ZydisMachineMode::LONG_COMPAT_32 instead of ZYDIS_MACHINE_MODE_LONG_COMPAT_32.

The final issue is whether you want your library to be accessed by C (and Objective-C, etc.) as well as C++.  In that case, you may do the following (which I just did in my local fork):

#ifdef __cplusplus
// The following is useful for debugging
#define ZYDIS_ENUM_DECL(SIZE, NAME) enum class NAME: SIZE
#define ZYDIS_ENUM_REF(ENUM, VALUE) ENUM::VALUE
#else
#define ZYDIS_ENUM_DECL(SIZE, NAME) typedef SIZE NAME; enum NAME##s
#define ZYDIS_ENUM_REF(ENUM, VALUE) VALUE
#endif

And then you declare your enum like so:

ZYDIS_ENUM_DECL(ZydisU8, ZydisMachineMode) {
    /**
     * @brief 64 bit mode.
     */
      LONG_64,
      …

And you refer to it like so:

        .. = ZYDIS_ENUM_REF(ZydisMachineMode, LONG_64);

flobernd:
We are aware of the struct size and I already tried to optimize it a bit. Your enum approach is not very practical for the following reasons:
- Usability (having to use a macro everytime I reference an enum value is not something I would like to do when working with an external library)
- ABI compatibility (the C enums will have different sizes than the C++ enums which can lead to problems, if compiled as dynamic library)

zlaski:
Valid concerns; however


  *   The enum-referencing macro (ZYDIS_ENUM_REF) would only be needed by code shared by C and C++.
  *   In v2, the stored enum values always have a size (e.g., ZydisU8), so there is no ABI discrepancy.  My perusal of your code on GitHub indicates that you are no longer constraining the size of enums (which will default to int). A C++ enum class without a size specifier should also default to int, so again, there will be no ABI inconsistency.  However, I would STRONGLY encourage you to constrain your enum sizes. 😊

flobernd:
> The enum-referencing macro (ZYDIS_ENUM_REF) would only be needed by code shared by C and C++

That's true, but not using the macros in a project depending on Zydis would then tag this project to a specific compiler (either C or C++).

> In v2, the stored enum values always have a size (e.g., ZydisU8), so there is no ABI discrepancy.

Yes, there is no problem with ABI compatibility in v2 with your code (and it really enhances the debuggability while using a C++ compiler), but - as you already discovered yourself - we switched to "real" enum types in v3 (to avoid the debugger problem you mentioned earlier aka. the debugger not showing enum values but plain ordinal numbers instead). I don't really see the benefit of C++ enums over the C ones, if we don't constrain the sizes anyways (and if we use constraints, we would either lose ABI compatibility or the better debuggability in C mode). Don't get me wrong ... strong typed enums (`enum class`) is a nice feature to be used in a pure C++ project, but I dont feel comfortable using such "hybrid approaches" in a project that was always designed to be a C project (a CMake switch could be an option, but even this does not feel very right).

I'm totally on your side that compacting the structs could have different benefits and at the same time I'm sad that pure C does not offer a standard way to deal with the enum sizes. We are already discussing other possibilites to shrink the struct size ... e.g. having a two-step API to first decode the instruction and speparate calls to later decode the operands. This way we would not have to define a fixed array of 10 operands (1070 bytes) in the `ZydisDecodedInstruction` struct.

PR #67, opened 2018-11-07
Update MSVC projects

Mattiwatti:
Hello hello, it's the MSVC guy again :smiley:

This PR updates the files under `/msvc` to make them up to date with the current master branch. I understand master is a moving target now that v3.0 is in development, so don't let this hold you back from breaking it all again in the next commit. But since the addition of the Zycore submodule is going to be the biggest change by far (project structure wise), I wanted to get that out of the way.

Refer to the first commit message for a summary of changes that should hopefully be noncontroversial. The other two commits fall under 'miscellaneous` so I separated them.

Some points to consider:
- The Zycore project is under `/msvc/dependencies/zycore`, which makes a lot of the paths in the project file `../../../look/like/this`. I decided that this was worth it because it preserves the property of `/msvc` that its layout mirrors the directory layout of the top level directory, which I would like to keep that way if possible.
- `ZycoreExportConfig.h` is located under `/msvc` (so next to `ZydisExportConfig.h`). I don't really have a preference for where to put this file and I can move it to `/msvc/dependencies` or `/msvc/dependencies/zycore` if needed. The current path was mainly chosen for two reasons:
  - It avoids having to add another include path to each project, because all projects already include `$(SolutionDir)`.
  - Hiding the file in a deep path might make it easy to overlook, despite being a pretty important header for the entire project.
- `ZYCORE_NO_LIBC` and `ZYAN_NO_LIBC` are both defined where applicable even though they mean the same thing. This should be changed to one of the two, pending [Zycore #1](https://github.com/zyantific/zycore-c/issues/1).
- The Debug DLL and Release DLL configurations currently produce both `Zycore.dll` and `Zydis.dll` (same behaviour as CMake if building shared libraries is enabled). Zydis does not link against or import from Zycore, but the tools and examples do potentially import from `Zycore.dll` (currently only `ZydisInfo` and `ZydisPerfTest` actually require it). In my opinion Zycore is small enough that producing a separate DLL for it is questionable. It may be worth considering making Zycore a static library even for 'DLL' configurations, so that executables continue to only depend on `Zydis.dll` but are free to use code from Zycore if needed.


flobernd:
Thank you very much for your PR :) We will merge it very soon.

1. I'm okay with the long relative pathes.
2. Placing these two configuration headers together sounds like a good idea for me. Woudn't like to have it hidden in a deep path either.
3. Oh, seems like I have to look at that again. Already cleaned up that mess a bit some time ago, but I missed the `LIBC` thing I guess.
4. The CMake file used to have a global option to define static/shared for all modules (and sub-modules). I did not like this solution for the same reasons you mentioned above and replaced it by a seperate `ZYXXX_SHARED_LIBRARY` option for each independent module (dll). This way you can choose to build `Zycore` as a static library and `Zydis` as a shared one, or the other way around, or both static, or both shared, and so on ... IMHO both of the major options (`Zycore shared + Zydis shared` as well as `Zycore static + Zydis shared`) are offering different advantages and disadvantages. Not really sure about this point at the moment, have to rething this.

Mattiwatti:
Re: 3: All actual code checks for `ZYAN_NO_LIBC`. `CMakeLists.txt` also uses `option(ZYAN_NO_LIBC, ...)`. The only mismatch is [here](https://github.com/zyantific/zycore-c/blob/ea3d80415af4691334f7bd62f8dc7cf34a2261ee/CMakeLists.txt#L131):
```
if (ZYCORE_NO_LIBC)
    target_compile_definitions("Zycore" PUBLIC "ZYCORE_NO_LIBC")
endif ()
```
where `ZYCORE_NO_LIBC` is used twice instead.

flobernd:
Haha great ... so apparently the `NO_LIBC` option never worked after my refactorings 😄 Thanks for reporting this. I'll fix that asap.

Mattiwatti:
OK, cool, that means the `ZYCORE_NO_LIBC` defines can be removed.

Do you know if a new commit can be merged on Github if I push changes to my branch after my PR has already been merged? If so, I'll push the change to my `msvc` branch ASAP. Otherwise I'll make a new PR (or you can make the change if you want: just delete all occurrences of `ZYCORE_NO_LIBC` in all `.vcxproj` files).

flobernd:
> Do you know if a new commit can be merged on Github if I push changes to my branch after my PR has already been merged?

Good question 😄 Not 100% sure, but you should be able to create a new PR on the existing branch, after you pushed something new.

PR #69, opened 2018-11-07
MSVC: remove unused ZYCORE_NO_LIBC defines

Mattiwatti:
Ref: #67

The answer to the question is: no, you must make a new PR :smile: But it can indeed be on the existing branch used for the previous one.


flobernd:
Thanks again :)

PR #70, opened 2018-11-08
ZydisWinKernel: remove leftover ZYDIS_SUCCESS()

Mattiwatti:
See title/commit message. I didn't catch this because I only compiled the release build (where `NT_ASSERT()` is a no-op).


athre0z:
Thanks!

PR #71, opened 2018-11-08
ZydisFormatterBasePrintDecorator: invert #if check for ZYAN_UNUSED()'s

Mattiwatti:
I came across this while compiling with `/WX` and testing the effects of disabling some features.

If `ZYDIS_DISABLE_AVX512` and `ZYDIS_DISABLE_KNC` are both defined, `ZYAN_UNUSED()` is needed here because otherwise the build will fail. That was probably the intent of the `#if` check, except it was testing the wrong way around.

### DIFF
--- a/src/FormatterBase.c
+++ b/src/FormatterBase.c
@@ -550,7 +550,7 @@ ZyanStatus ZydisFormatterBasePrintDecorator(const ZydisFormatter* formatter,
     ZYAN_ASSERT(buffer);
     ZYAN_ASSERT(context);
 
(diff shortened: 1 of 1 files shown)


athre0z:
Thanks!

I think at one point the flags had negated names (ZYDIS_ENABLE_XXX) and we forgot to invert that when refactoring it to be `ZYAN_DISABLE_XXX` instead (to have sane defaults for people who just C&P our stuff into their project instead of using CMake).

PR #94, opened 2019-02-19
Update MSVC projects

### LINKED ISSUE #91: Set Zydis include paths as project-relative, not solution-relative
It is necessary when you want to add Zydis project as subproject to another solution without Zydis.vcxproj modification (when it presents in project as git submodule)

Mattiwatti:
This PR mostly just fixes #91, which was less work than I expected. I also decided to finally add `Formatter01`-`Formatter03` since they still haven't been given proper names yet :stuck_out_tongue: 

Note w.r.t. build output: this now goes to `bin/<arch>` instead of `bin` (see the issue for my comment re: this). It's what was requested in the issue and I'm fine with either. Will revert this part if anyone has strong opposition for some reason.


flobernd:
Thanks for the PR! Outputting to `bin/<arch>` is fine for me. What do you think @athre0z?

athre0z:
No real preference -- if some users prefer it this way, it's fine with me.

flobernd:
Cool, merged. Thanks again!

PR #95, opened 2019-02-19
Fix signed vs unsigned comparisons

Mattiwatti:
These come from Clang's `-clang-diagnostic-sign-compare`. I doubt any actual bugs were fixed by this, but it removes a bunch of exclamation marks in my IDE.


athre0z:
Thanks!

PR #96, opened 2019-02-19
ZydisWinKernel fixes

Mattiwatti:
- Fix NULL comparison against wrong variable
- Cast some printf argument types to match the format specifier

The second one was found by Resharper after decorating `Print()` with `[[gnu::format(printf, fmtIndex, firstToCheck)]]` (there is a bit more boilerplate than that, but that's the basic idea). I could open a bug in Zycore to add this to `Defines.h` so that e.g. `ZYAN_PRINTF()` can also get this argument type checking and highlighting?

printf_attr

### DIFF
--- a/examples/ZydisWinKernel.c
+++ b/examples/ZydisWinKernel.c
@@ -114,13 +114,13 @@ DriverEntry(
     if (imageBase == 0)
         return STATUS_DRIVER_ENTRYPOINT_NOT_FOUND;
     const PIMAGE_NT_HEADERS ntHeaders = RtlImageNtHeader((PVOID)imageBase);
-    if (imageBase == 0)
+    if (ntHeaders == NULL)
         return STATUS_INVALID_IMAGE_FORMAT;
 
     // Get the section headers of the INIT section
     PIMAGE_SECTION_HEADER section = IMAGE_FIRST_SECTION(ntHeaders);
     PIMAGE_SECTION_HEADER initSection = NULL;
-    for (ULONG i = 0; i < ntHeaders->FileHeader.NumberOfSections; ++i)
+    for (USHORT i = 0; i < ntHeaders->FileHeader.NumberOfSections; ++i)
     {
         if (memcmp(section->Name, "INIT", sizeof("INIT") - 1) == 0)
         {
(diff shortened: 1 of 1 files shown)


flobernd:
Thanks, will merge this later.

I just checked the ReSharper (ReSharper C++ 2018.3.3) argument type checking and highlighting for `ZYAN_PRINTF()` and it seems to work for me without adding an extra decorator.

Mattiwatti:
Ah, yes, I'm guessing you were checking from a project without `ZYAN_NO_LIBC` defined? In that case `ZYAN_PRINTF` equals `printf` and Resharper 'gets' it because `printf` is part of the stdlib. (With `ZYAN_NO_LIBC` defined there is no `ZYAN_PRINTF`, which makes sense...). `ZyanStringInsertFormat` and `ZyanStringAppendFormat` could still benefit from it though. Don't know if that makes it worth it.

flobernd:
I'm not familar with these decorators. How do they perform in terms of compatibility with different compilers and how much work is required to add them? I think supporting this feature would definitely not hurt, if there are no drawbacks.

Mattiwatti:
The only 'compilers' (Resharper is more like a parser I guess) that support it are GCC and Resharper as far as I know. But you can simply define the attribute to a blank for other compilers. Resharper understands the GNU syntax for `printf` and additionally supports a `wprintf` attribute for wide strings with its own `rscpp` attribute. Boilerplate code I got from the Resharper bugtracker:
```C
// One-time defines to be added in some common header
// NB: indices are 1-based. firstToCheck = 0 can be used for va_list taking functions
#if defined(__RESHARPER__) || defined(__GNUC__)
#define PRINTF_ATTR(fmtIndex, firstToCheck) \
    [[gnu::format(printf, fmtIndex, firstToCheck)]]
#else
#define PRINTF_ATTR(fmtIndex, firstToCheck)
#endif

#if defined(__RESHARPER__)
#define WPRINTF_ATTR(fmtIndex, firstToCheck) \
    [[rscpp::format(wprintf, fmtIndex, firstToCheck)]]
#else
#define WPRINTF_ATTR(fmtIndex, firstToCheck)
#endif
```
Example usage:
```C
PRINTF_ATTR(2, 3)
void
MyErrorPrintf(
    int errorLevel,
    const char* Format,
    ...
    );
```
This will give proper highlighting and type checking for all calls to `MyErrorPrintf()`.

flobernd:
Looks cool. Are you going to send us a pull-request in the zycore-c repo for the highlighting stuff?

Merged the ZydisWinKernel fixes now .. Thanks again!

Mattiwatti:
Sure, will do.

Issue #46, opened 2018-07-11
-fPIC missing on FreeBSD


gdbinit:
Hi,

FreeBSD clang doesn't seem to like the built libZydis.a due to missing -fPIC option while compiling.
/usr/bin/ld: /usr/local/lib/libZydis.a(Decoder.c.o): relocation R_X86_64_32S against `a local symbol' can not be used when making a shared object; recompile with -fPIC
/usr/local/lib/libZydis.a: could not read symbols: Bad value
clang: error: linker command failed with exit code 1 (use -v to see invocation)

Fixing it by adding 
target_compile_options("${target}" PRIVATE "-fPIC")
to function (_set_common_flags target) in CMakeLists.txt.

It seems FreeBSD specific issue so maybe detect FreeBSD and add the flag to compiler options?

$ uname -an
FreeBSD freebsd 11.2-RELEASE FreeBSD 11.2-RELEASE #0 r335510: Fri Jun 22 04:32:14 UTC 2018     root@releng2.nyi.freebsd.org:/usr/obj/usr/src/sys/GENERIC  amd64
$ clang -v
FreeBSD clang version 6.0.0 (tags/RELEASE_600/final 326565) (based on LLVM 6.0.0)
Target: x86_64-unknown-freebsd11.2
Thread model: posix
InstalledDir: /usr/bin

athre0z:
Hmm, any chance that might be due to interference with other flags in your project or something? Seems to be working fine in my test run.

```
ath@fbsd-vm ~ % git clone https://github.com/zyantific/zydis.git
Cloning into 'zydis'...
remote: Counting objects: 4675, done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 4675 (delta 10), reused 15 (delta 4), pack-reused 4641
Receiving objects: 100% (4675/4675), 9.10 MiB | 6.21 MiB/s, done.
Resolving deltas: 100% (3244/3244), done.
ath@fbsd-vm ~ % cd zydis
ath@fbsd-vm ~/zydis (git)-[master] % mkdir build && cd build
ath@fbsd-vm ~/zydis/build (git)-[master] % cmake -DZYDIS_BUILD_EXAMPLES=OFF ..
-- The C compiler identification is Clang 6.0.0
-- The CXX compiler identification is Clang 6.0.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
[...]

PR #102, opened 2019-03-06
add -fPIC option for gcc

sincoder:
make zydis easy used in shared object

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,12 @@ include(GNUInstallDirs)
 
 project(Zydis VERSION 3.0.0.0 LANGUAGES C CXX)
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+SET(GCC_COVERAGE_COMPILE_FLAGS "-fPIC")
+SET(GCC_COVERAGE_LINK_FLAGS    "")
+add_definitions(${GCC_COVERAGE_COMPILE_FLAGS})
+endif()
+
 # =============================================================================================== #
 # Overridable options                                                                             #
 # =============================================================================================== #


flobernd:
CMake should automatically set the `-fPIC` parameter, if you use the `ZYDIS_BUILD_SHARED_LIB` option. It did for me at least ..

sincoder:
@flobernd  But it only create libZydis.so file , if our shard object want static link to libZydis.a , still not possible

flobernd:
Ah, I understand .. so you have a shared library project that you want to link against a static Zydis lib.

Let us check, if there are any drawbacks with forcing position independent code. If not, we can probably add a CMake option for your PR.

athre0z:
CMake has an [`CMAKE_POSITION_INDEPENDENT_CODE`](https://cmake.org/cmake/help/v3.14/variable/CMAKE_POSITION_INDEPENDENT_CODE.html) option to force PIC for all targets. Also check [this response](https://github.com/zyantific/zydis/issues/46#issuecomment-404314617) in a similar issue.

flobernd:
Are you using FreeBSD as well @sincoder?

flobernd:
We decided to not generally force PIC for all build configurations, as it might have negative impact on some platforms and might not be supported on other ones.

As a workaround we suggest you to set the `POSITION_INDEPENDENT_CODE ` property in your projects CMakeFile.txt for both Zycore and Zydis:
```cmake
target_link_libraries("MyProject" PUBLIC "Zydis")

...

set_property(TARGET "Zycore" PROPERTY POSITION_INDEPENDENT_CODE ON)
set_property(TARGET "Zydis" PROPERTY POSITION_INDEPENDENT_CODE ON)
```

jessielesbian:
hi!

PR #103, opened 2019-03-19
MSVC: set Windows SDK to whatever the highest available version is

Mattiwatti:
Hiya,

I had a small issue with MSVC after pulling 38df86f from upstream. I'm fine with the project changes, but VS 2017 does not recognize '10.0' as a valid SDK version, and I need to set the SDK to a specific version to get the projects to compile. As far as I know all 10.x SDKs have specific version numbers like 10.0.17763.0, but maybe '10.0' is a newly accepted value in the VS 2019 beta? Not sure. I hope so, that would mean MS is finally doing some sort of semver on these SDK releases which are never breaking anyway.

I always thought I fixed this (not historically a problem, but it is now that Windows gets a new SDK every 6 months) in my projects by setting the value of `WindowsTargetPlatformVersion` to `$(LatestTargetPlatformVersion)`, but it turns out this only works for driver projects that use the WDK. Oops.

After some digging I found this (admittedly undocumented) value that will set the SDK to whatever the highest local 10.0.X version is. This is something that CMake normally does when generating the projects. Because `/msvc` is under source control, this dumb issue causes all the project files to be in the change list for no good reason.

### DIFF
--- a/msvc/dependencies/zycore/Zycore.vcxproj
+++ b/msvc/dependencies/zycore/Zycore.vcxproj
@@ -89,7 +89,7 @@
     <Platform Condition="'$(Platform)' == ''">x64</Platform>
     <RootNamespace>$(MSBuildProjectName)</RootNamespace>
     <Keyword>Win32Proj</Keyword>
-    <WindowsTargetPlatformVersion>10.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>$([Microsoft.Build.Utilities.ToolLocationHelper]::GetLatestSDKTargetPlatformVersion('Windows', '10.0'))</WindowsTargetPlatformVersion>
   </PropertyGroup>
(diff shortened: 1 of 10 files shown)


flobernd:
Hi there,

thanks for your PR. It is really annoying that VS does not provide a simple way to specify the highest available toolset and Windows SDK :/ Same issue forced me downgrading the toolset to make the project compatible with VS17 again .. and now it's ofc not compatible with VS19 anymore (if you dont have the 'old' toolset installed).

Nice find with the undocumented workaround!

Mattiwatti:
For the platform toolset, I believe `<PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>` should work. E.g. if you have both VS 2017 and 2019 installed, the selected toolset will be v141 or v142 respectively, depending on which version you opened the project with.

Want me to make a PR for this?

PR #116, opened 2019-04-09
Use canonical include format

lars-t-hansen:
By way of background, we're trying to incorporate zydis in Firefox development builds to disassemble the code for our new wasm jit backend, see [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1539806).

While rewriting the zydis sources slightly to fit into our build system I noticed that one of the #include directives uses the quoted form instead of angle-bracketed form; this looks like an oversight to me.  The attached patch fixes this.


athre0z:
Thanks again!

PR #124, opened 2019-08-24
Port Azure CI to GitHub actions

athre0z:
Please apologize the mail flood -- there is currently no way to test these on private repos.


athre0z:
~20 force-pushes and a few workarounds for bugs in the job matrix later, everything is now working fine. @flobernd what do you think, should we throw out Azure despite GitHub actions still being in beta? Given the frequency at that we receive PRs from external contributors, I think that might be acceptable. If everything goes south, we can still test PRs manually.

flobernd:
Yes, let's switch to GitHub actions and throw out Azure!

athre0z:
aaand it's gone!

PR #135, opened 2020-02-11
Allow zycore dep path to be changed via CMake var

athre0z:
This allows for flattening dependencies and splitting Zydis and Zycore into separate bundles in package managers. This is required to make the vcpkg people happy in regards to their wish to split Zycore/Zydis into two packages.

I'm pretty sure we had such a flag in previous versions -- do you happen to remember why we removed it? Perhaps my memory is off and this was in another project. Anyways: do you see any reason not to support that?

If you're ok with this, please just merge via rebase.


flobernd:
I think we always wanted to add this as an option but we never did until now 😄

Issue #137, opened 2020-03-11
Provide CMake package config file


RIscRIpt:
CMake can generate [package config file](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html), which can be used in other projects that use CMake.

This way other projects can simply write `find_package(zydis)`. If it's not installed in the system, users can provide path to `zydis-config.cmake` with `-Dzydis_DIR`.

PR #138, opened 2020-03-11
[#137] Add CMake package config file

RIscRIpt:


athre0z:
Thanks! I've always been too lazy to find out how to make it work -- looks like I won't have to now!

PR #142, opened 2020-05-07
Make file list non-empty in m.css-generated docs

mosra:
Original issue: https://github.com/mosra/m.css/issues/126

Makes the file list show up correctly in the docs:

image

**Disclaimer:** The patch is just the minimal set of changes needed to achieve this. I didn't follow any coding guidelines and didn't know where to put the new file either (those doc blocks could be directly in the include files as well, for example). Feel free to take any of this or ignore the changes altogether :)

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -791,7 +791,8 @@ WARN_LOGFILE           =
 # Note: If this tag is empty the current directory is searched.
 
 INPUT                  = ./include \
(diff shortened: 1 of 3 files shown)


athre0z:
Thanks so much! <3

I'll just merge this as-is and then later try to integrate this into the file headers.

PR #156, opened 2020-09-25
Can you guys please change it to pointer-oriented? thanks!

jessielesbian:


flobernd:
This is a C library. The new keyword is only available for C++.

athre0z:
Even if it wasn't for the fact that `new` is a C++ keyword, this just makes the example worse because it unnecessarily heap allocs (when one of the core values of Zydis is to not have to heap-alloc) and additionally also leaks memory because you didn't free the pointers in the end.

PR #158, opened 2020-09-26
Fix: zydis_decoder_tree_root multiple definitions

tathanhdinh:
`zydis_decoder_tree_root` is redefined in the header file, as https://github.com/zyantific/zydis/blob/99ca4c62c94201a66c2afa7dda75865d963b083e/include/Zydis/Internal/DecoderData.h#L290
while it shouldn't be.


athre0z:
Thanks!

Issue #76, opened 2018-11-13
fatal error: Zycore/Defines.h: No such file or directory


qiuyuX:
When I compile the example code, the following missing header issue appears:
/usr/local/include/Zydis/Zydis.h:35:28: fatal error: Zycore/Defines.h: No such file or directory
/usr/local/include/Zycore/String.h:35:32: fatal error: ZycoreExportConfig.h: No such file or directory
/usr/local/include/Zydis/MetaInfo.h:35:31: fatal error: ZydisExportConfig.h: No such file or directory

It seems the required headers are not installed by running "make install".

Also a side issue, the example in https://zyantific.com/doc/zydis/ seems outdated. Initially I followed this example and got errors like "undefined reference to `ZYDIS_SUCCESS'".

My machine is ubuntu16.04.

athre0z:
The examples should be built automatically as part of the CMake build progress. If you want to use manual linkage using regular unix makefiles, you'll have to add a few extra include directories that are normally automatically added by CMake, namely 
```-I../zydis/dependencies/zycore/include -I../zydis/include -I../zydis/build -I../zydis/build/dependencies/zycore``` (assuming that your project lives in a directory on the same level where Zydis lives). I'd generally recommend using CMake for your project as well, since it is the de-facto standard for C and C++ cross platform stuff these days.

`master` is the current in-dev version (bleeding edge code) of `v3`, the documentation for that is available [here](https://zydis.re/doc/3/) (the link can be found in the README.md of this repo). The documentation your URL refers to is for the current stable version `v2` (which can be found on the `maintenance/v2` branch). For anything except for testing and playing around, I recommend going with the stable version.

athre0z:
Nevertheless, `make install` should indeed install the `zycore` files as well. I'll have to check on this ...

qiuyuX:
Thanks for your explanation! I will try my code based on the maintenance branch.

Below is the output when I ran "make install":
[ 19%] Built target Zycore
[ 61%] Built target Zydis
[ 66%] Built target Formatter03
[ 72%] Built target Formatter02
[ 77%] Built target ZydisFuzzIn
[ 83%] Built target ZydisInfo
[ 88%] Built target Formatter01
[ 94%] Built target ZydisDisasm
[100%] Built target ZydisPerfTest
Install the project...
-- Install configuration: ""
-- Up-to-date: /usr/local/lib/libZydis.a
-- Installing: /usr/local/include/ZydisExportConfig.h
[...]

PR #160, opened 2020-10-11
Install zycore

Clcanny:
If we don't install zycore, compiler will complain: 

```cpp
#include <Zydis/Zydis.h>
int main() {
}
```

```bash
# g++ -std=c++11 test_zydis.cpp -lZydis -O0 -ggdb -o test_zydis
In file included from test_zydis.cpp:1:0:
/usr/local/include/Zydis/Zydis.h:35:28: fatal error: Zycore/Defines.h: No such file or directory
 #include <Zycore/Defines.h>
                            ^
compilation terminated.
```

Test environment:

```bash
# uname -r
4.9.0-9-amd64
# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
# cmake --version
cmake version 3.7.2
# g++ --version
g++ (Debian 6.3.0-18+deb9u1) 6.3.0 20170516
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,7 @@ if (NOT EXISTS "${ZYDIS_ZYCORE_PATH}/CMakeLists.txt")
     )
 endif ()
 
-add_subdirectory(${ZYDIS_ZYCORE_PATH} "zycore" EXCLUDE_FROM_ALL)
+add_subdirectory(${ZYDIS_ZYCORE_PATH} "zycore")
 
 # =============================================================================================== #
 # Library configuration                                                                           #


Clcanny:
@athre0z Hello, please forget my poor English. Please take some time to review this change.
And I don't know why EXCLUDE_FROM_ALL is needed in add_subdirectory statement.
Should I only install header files of zycore?

Clcanny:
@athre0z Hello, please forget my poor English.Please take some time to review this change.
P.S. I don't know why EXCLUDE_FROM_ALL is needed in add_subdirectory statement.
Should I only install header files of zycore?

flobernd:
Hi @Clcanny,

Is this an installation specific case? I tested Zydis on Debian multiple times and never had problems with the CMake project.

The EXCLUDE_FROM_ALL flag should just exclude the Zycore files from the generated project, but should not prevent the compiler from finding the headers.

Maybe we are missing something here in case of CMake install.

Clcanny:
```Dockerfile
FROM debian:buster

ENV DEBIAN_FRONTEND noninteractive
RUN apt-get update
RUN apt-get install -y build-essential make cmake gcc g++ git

# Compile zydis.
WORKDIR /root
RUN git clone --recursive https://github.com/zyantific/zydis.git
WORKDIR /root/zydis
RUN mkdir build && cd build && cmake .. && make && make install

# Compile source file.
RUN echo "#include <Zydis/Zydis.h>\nint main {}" > test_zydis.cpp
RUN g++ -std=c++11 test_zydis.cpp -lZydis -O0 -ggdb -o test_zydis

CMD /bin/bash
```

Hi, @athre0z , this is a example of how i use cmake install.

flobernd:
Alright that's what I wanted to know. The normal build works fine, but the installation seems to miss on the `Zycore` module. Sadly I don't think your change is the correct way to fix that.

What do you think @athre0z ?

Clcanny:
> Alright that's what I wanted to know. The normal build works fine, but the installation seems to miss on the `Zycore` module. Sadly I don't think your change is the correct way to fix that.
> 
> What do you think @athre0z ?

Maybe just copy files of zycore?

athre0z:
Hi @Clcanny! Thanks for the PR and the Dockerfile for reproduction. I think this is the same issue as in #76. The problem with the approach of installing Zycore as part of Zydis is that if you have another lib that depends on Zydis and Zycore (e.g. our hooking lib), you'd kind of install it twice.

The way that it's done in e.g. vcpkg is to just bulid Zycore individually -- applying this to your Dockerfile:

```dockerfile
FROM debian:buster

ENV DEBIAN_FRONTEND noninteractive
RUN apt-get update
RUN apt-get install -y build-essential make cmake gcc g++ git

# Compile zydis.
WORKDIR /root
RUN git clone --recursive https://github.com/zyantific/zydis.git
WORKDIR /root/zydis/dependencies/zycore
RUN mkdir build && cd build && cmake .. && make && make install
WORKDIR /root/zydis
RUN mkdir build && cd build && cmake .. && make && make install

# Compile source file.
RUN echo "#include <Zydis/Zydis.h>\nint main() {}" > test_zydis.cpp
RUN g++ -std=c++11 test_zydis.cpp -lZydis -O0 -ggdb -o test_zydis

CMD /bin/bash
```

I think I'd prefer keeping things this way, but am also happy to further discuss this if different opinions on this come up.

Clcanny:
@flobernd @athre0z Thanks a lot for your reply! Now I know why we don't install Zycore. And I will close this pull request.

PR #162, opened 2020-11-14
Regression tests

flobernd:
Add basic regression test script and some cases to start with.


tests/regression.py
```diff
@@ -0,0 +1,57 @@
...
+
+parser = argparse.ArgumentParser(description="Regression testing.")
+parser.add_argument(dest="operation", choices=["test", "rebase"])
+parser.add_argument(dest="zydis_info_path", type=str)
+args = parser.parse_args()
+
+has_failed = False
+
+for case in os.listdir(TEST_CASE_DIRECTORY):
+    if case.endswith(".in"):
```
athre0z, line 29: Early-exit would reduce indentation here.

tests/regression.py
```diff
@@ -0,0 +1,57 @@
...
+            continue
+
+        try:
+            with open(path, mode="rb") as f:
+                s = f.read()
+
+            if s != out:
+                print(f"FAILED: '{case}' [{payload}]")
+                has_failed = True
+        except:
```
athre0z, line 53: I think this should be a bit more specific rather than eating up all errors.

```suggestion
        except FileNotFoundError:
```

athre0z (approved):
LGTM otherwise!

athre0z (approved):

PR #163, opened 2020-11-18
Add unofficial Haskell binding to readme

nerded1337:
Hello guys!

I am the owner of [this haskell binding](https://github.com/nerded1337/zydiskell) and would like to have it  listed despite the fact that we don't provide everything yet (only instruction decoding, no string formatting as of now).

The package will be available on [Hackage](https://hackage.haskell.org/package/zydiskell-0.1.0.1) as well.

What do you think :)?


flobernd:
Looks good to me :) Bindings are always very much appreciated. Thank you!

PR #170, opened 2021-02-27
Refactor `accessed_flags` for more intuitive and performant acccess

flobernd:


src/Decoder.c
```diff
@@ -4804,8 +4804,12 @@ static ZyanStatus ZydisDecodeInstruction(ZydisDecoderContext* context,
...
                             sizeof(flags->action));
 
-                        instruction->flags_read = flags->flags_read;
-                        instruction->flags_written = flags->flags_written;
+                        instruction->cpu_flags_read = flags->flags_read & 0x003FFFFF;
+                        instruction->cpu_flags_written = flags->flags_written & 0x003FFFFF;
+                        instruction->fpu_flags_read = 
+                            (ZydisFPUFlags)((flags->flags_read & 0x03C00000) >> 22);
+                        instruction->fpu_flags_written =
+                            (ZydisFPUFlags)((flags->flags_written & 0x03C00000) >> 22);
```
athre0z, line 4812: I'm a bit confused: what constant is going to correspond to that after the shift? `ZYDIS_CPUFLAG_C0` is defined as `22`, so it doesn't appear to be expecting a shift.

src/Decoder.c
```diff
@@ -4804,8 +4804,12 @@ static ZyanStatus ZydisDecodeInstruction(ZydisDecoderContext* context,
                         ZYAN_MEMCPY(&instruction->accessed_flags, &flags->action,
                             sizeof(flags->action));
 
-                        instruction->flags_read = flags->flags_read;
-                        instruction->flags_written = flags->flags_written;
+                        instruction->cpu_flags_read = flags->flags_read & 0x003FFFFF;
```
athre0z, line 4807: Not a fan of the hardcoded mask. Can we do `(1 << MAX_FLAG) - 1` here?

src/Utils.c
```diff
@@ -190,29 +190,25 @@ ZyanStatus ZydisGetAccessedFlagsByAction(const ZydisDecodedInstruction* instruct
 ZyanStatus ZydisGetAccessedFlagsRead(const ZydisDecodedInstruction* instruction,
```
athre0z, line 190: Should we mark these functions as deprecated, given how trivial they are now?

athre0z (commented):
Looks good in general, I added a few comments. :)

include/Zydis/Internal/SharedData.h
```diff
@@ -879,8 +879,8 @@ typedef struct ZydisInstructionDefinitionMVEX_
 typedef struct ZydisAccessedFlags_
 {
     ZydisCPUFlagAction action[ZYDIS_CPUFLAG_MAX_VALUE + 1];
-    ZyanU32 flags_read      ZYAN_BITFIELD(22);
```
athre0z, line 882: Given that these are bitfields anyways, can't we just split out the FPU flags here as well, then get rid of the manual masks in `Decoder.c` entirely, leaving it to the compiler to deal with?

include/Zydis/DecoderTypes.h
```diff
@@ -435,104 +435,115 @@ typedef ZyanU64 ZydisInstructionAttributes;
...
-} ZydisCPUFlag;
+/**
+ * Defines the `ZydisCPUFlag` data-type.
+ */
+typedef ZyanU8 ZydisCPUFlag;
+
+/**
+ * Carry flag.
+ */
+#define ZYDIS_CPUFLAG_CF     0
```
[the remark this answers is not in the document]
athre0z: Ok, I remembered that we also do this for all other things that are flags. However, the other flags (e.g instruction attributes) generally use the mask rather than the shift count as the value (e.g. `0x10` instead of 4). Should we also do this here?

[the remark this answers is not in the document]
flobernd: That as well came to my mind, but this is the only way to be backwards compatible.

I don't really mind having a breaking release this time tho. Effort to fix existing implementations is very minor for the users and at some point we have to remove the deprecated stuff anyways.

include/Zydis/Internal/SharedData.h
flobernd: Yes, already planned. Requires some additional generator changes so I did it quick&dirty for now.

Will improve this before we merge the PR.

src/Decoder.c
flobernd: This mask will be obsolete after doing the generator change.

src/Decoder.c
flobernd: You are right, for now there are no corresponding defines (the CPUFLAG ones are for backwards compatibility - e.g. to use as an index in the accessed_flags array). I will create additional FPUFLAG ones for improved usability. The CPUFLAG ones (only C0..C3) will be removed in the breaking release.

include/Zydis/DecoderTypes.h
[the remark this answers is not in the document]
athre0z: Ah yeah, I see. Hmmm, that is unfortunate. Still, I don't want to break compat. We promise semantic versioning in the README and we should stick to that. We could release it as 4.0, but it kind of seems overkill. I'd say leave it as is.

flobernd:
Let me do some small modifications to the `ZydisInfo` tool before we merge it tomorrow.

flobernd:
@athre0z Feel free to complete after a final review.

athre0z (approved):
LGTM!

PR #191, opened 2021-08-12
Fix cmake config files

abouvier:
The current `config.cmake` file is almost empty and is installed in the wrong directory.
Also add the `config-version.cmake` file.
Similar patch from https://github.com/zyantific/zycore-c/pull/27 is needed.

This will allow other projects to do:

```
find_package(Zydis REQUIRED)
target_link_libraries(project PRIVATE Zydis::Zydis)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,7 +91,10 @@ endif ()
 
 target_link_libraries("Zydis" PUBLIC "Zycore")
 target_include_directories("Zydis"
-    PUBLIC "include" ${PROJECT_BINARY_DIR}
+    PUBLIC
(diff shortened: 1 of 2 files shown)


flobernd:
Hey @abouvier, thanks for your contribution! @athre0z will review the changes within the next days. Looks nice from the first sight.

flobernd:
Note to myself/@athre0z - don't forget to update the `Zycore` submodule after merging both PRs.

athre0z:
I updated the submodule on @abouvier's branch now (and squashed it into his commit) so that we can see whether build tests pass before merging. Since they do: merged! :-) Thanks for the PR!

PR #193, opened 2021-08-14
Fixes #192

### LINKED ISSUE #192: Incorrect formatting of signed displacements
All displacements are treated as signed and that works fine in most cases except for absolute addresses in relative mode (`ZYDIS_RUNTIME_ADDRESS_NONE`). This mode breaks for instructions with displacement-only SIB bytes and moffs operands. This can easily lead to some very confusing outputs, especially for Intel syntax.

Examples:
```
ZydisInfo.exe -32 67a08888
...
== [      ATT ] ============================================================================================
   ABSOLUTE: mov 0x00008888, %al
   RELATIVE: mov -0x7778, %al

== [    INTEL ] ============================================================================================
   ABSOLUTE: mov al, byte ptr ds:[0x00008888]
   RELATIVE: mov al, byte ptr ds:[0x7778]

ZydisInfo.exe -32 67a07877
== [      ATT ] ============================================================================================
   ABSOLUTE: mov 0x00007778, %al
   RELATIVE: mov 0x7778, %al

== [    INTEL ] ============================================================================================
   ABSOLUTE: mov al, byte ptr ds:[0x00007778]
   RELATIVE: mov al, byte ptr ds:[0x7778]
```
Relative mode should apply only to displacements that are actually relative to something other than segment base i.e. explicit register(s) or RIP.

Seems like a small fix is required in code of `ZydisFormatterXXXFormatOperandMEM` functions to correct this. Example fix for Intel (it's practically the same for AT&T):
```diff
-        if (context->operand->mem.disp.has_displacement && (context->operand->mem.disp.value
-            || neither_reg_nor_idx))
+        if (neither_reg_nor_idx)
+        {
+            ZYAN_CHECK(formatter->func_print_address_abs(formatter, buffer, context));
+        }
+        else if (context->operand->mem.disp.has_displacement && context->operand->mem.disp.value)
         {
             ZYAN_CHECK(formatter->func_print_disp(formatter, buffer, context));
         }
```
By the way I found this while investigating some issues I had with signedness of immediates and other values in Zydis. For example why in `mov al, moffs8 (A0)` source operand is treated as unsigned (`uint8`) but in `mov ax/eax/rax, moffs16/32/64 (A1)` element type is specified as `int`? I know that defining signedness in data transfer instructions (except for zero/sign-extending variants of mov) is pointless but I would expect some consistency here (all signed, all unsigned or something like "not applicable").

mappzor:
Fixed #192 + some quality of life improvements for Windows users in regression.py


flobernd (approved):
Looks good from the first sight.

tests/regression.py
```diff
@@ -1,3 +1,4 @@
+#! python3
```
bjorn3, line 1: This is incorrect on unix. "python3" is not an absolute path. It would have to be something like

```suggestion
#!/usr/bin/env python3
```

bjorn3:
Did you mean to remove src/Generated/EncodableInstructions.inc?

flobernd:
> 
> 
> Did you mean to remove src/Generated/EncodableInstructions.inc?

That's fine from my side as it's unused anyways. Thanks for mentioning.

tests/regression.py
flobernd: Thanks for pointing that out. The GitHub Actions workflow does not seem to have a problem with it, but it's always better to be on the safe side.

tests/cases/default_021.out
```diff
@@ -1,36 +1,36 @@
...
-   RELATIVE: mov %eax, 0xF887AE9C6A8107B
-
-== [    INTEL ] ============================================================================================
-   ABSOLUTE: mov dword ptr ds:[0x0F887AE9C6A8107B], eax
-   RELATIVE: mov dword ptr ds:[0xF887AE9C6A8107B], eax
-
-== [ SEGMENTS ] ============================================================================================
-A3 7B 10 A8 C6 E9 7A 88 0F 
-:  :..DISP
-:..OPCODE
```
athre0z, line 36: It seems like you have somehow committed this (as well as all other new test cases) with CRLF newlines. Usually, line endings [should be normalized to LF on commit by git](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings).

athre0z (commented):
LGTM otherwise -- thanks for the fix! :)

tests/cases/default_021.out
mappzor: autocrlf/safecrlf comes with its own set of issues and I have it turned off globally. Typically I handle EOL with `.gitattributes` but it's not used here, so this preference has bitten me here ;)

athre0z (approved):

PR #194, opened 2021-08-16
Enhanced formatting for MIB operands

mappzor:
Although index register doesn't participate in address calculation, it still encodes an operand (pointer value) and should be printed by the formatter (scale should be ignored completely though). Without this several encodings with different semantics become ambiguous. This PR affects `bndldx` and `bndstx` instructions only. New formatting follows style used in Intel's manual (Vol. 1 Chapter 17: Memory Protection Extensions).

Ignoring scale could be acheived by changes in decoder but since that (useless) information is technically encoded in the instruction it's probably better to just change formatters and leave decoding as it is.

By the way giving user an option to skip printing scale == 1 might be worth considering as a future enhancement.


flobernd:
Looks good to me! Thanks again for the contribution. Back then the style was different I guess, otherwise I can not explain why I should have added an explicit condition to not print the index register for `MIB` operands.

> By the way giving user an option to skip printing scale == 1 might be worth considering as a future enhancement.

Good suggestion. I will create an issue for that 😄

flobernd (approved):

PR #199, opened 2021-08-24
Improved handling of 16-bit relative operands (Fixes #198)

### LINKED ISSUE #198: Improper handling of 16-bit relative operands in ZydisCalcAbsoluteAddress
Relative operands are typically sign-extended and added to EIP/RIP but in case of 16-bit operands there's an additional truncating behavior (address is masked with 0xFFFF). `ZydisCalcAbsoluteAddress` attempts to handle this special case with `operand->size == 16` check but unfortunately this isn't sufficient. This check uses physical size of passed operand instead of effective operand size for the whole instruction. This handles only some cases (e.g. AMD64-specific `rel16` branches or `call` instruction in 32-bit mode with 66 prefix) but fails in others (e.g. 66 prefix on `rel8` branches).

Fix seems trivial but to avoid regressions I decided to examine all instructions with relative operands as there aren't too many of them. All following regular control flow instructions share mentioned truncating behavior:
```
call loop loope loopne jmp jcxz jecxz jrcxz jz jnz jb jnb jbe jnbe jl jnl jle jnle jo jno jp jnp js jns
```
Instructions specific to Xeon Phi (`jkzd jknzd`) also behave in the same way. However "x" in x86 stands for eXception, so of course there's a single instruction that breaks seemingly well-established rules. It's the `xbegin` instruction with following remark in Intel SDM:
>The instruction also specifies a relative offset to compute the address of the fallback code path following a transactional abort. (Use of the 16-bit operand size does not cause this address to be truncated to 16 bits, unlike a near jump to a relative offset.)

This is a special case and has to be hardcoded.

mappzor:


athre0z (commented):
Very good catch & thanks for the fix, excellent work!

We usually try not hardcoding conditions on instructions and introduce attributes for cases like this, but I'll admit that this would probably be quite overkill if it affects just a single instruction. There are a selection of other such cases, so I think it's probably fine for this. I'd love if you could paste the portion that you quoted from the SDM (in #198) into a comment that explains this special case, so we remember why it's there later. :-)

flobernd:
Thanks a lot! Merged.

Issue #200, opened 2021-08-26
Zydis does not recognize Intel CET NOTRACK prefix


tremalrik:
The Intel CET instruction set extension defines a "notrack" prefix that can be used to disable indirect branch tracking for individual indirect-branch instructions even when CET is enabled. The prefix is 3E, same as the DS segment-override prefix; as such, an opcode like e.g. 3E FF D0 in 64-bit mode should decode to something like "notrack call rax" rather than just the "call rax" that Zydis returns now.

The exact rules for when the prefix is recognized as a "notrack" prefix is given by the Intel SDM (volume 1, section 18.3.1):
> In 64-bit mode, the 3EH  prefix on an indirect CALL or JMP is recognized as a no-track prefix if there isn't a 64H/65H prefix on the instruction. 
> In legacy/compatibility mode, the 3Eh prefix on an indirect CALL or JMP is recognized as a no-track prefix when it is the last group 2 prefix on the instruction.

This would seem to imply that an instruction encoding like e.g. 26 3E 2E FF D0 (call eax/rax with ES,DS and CS segment prefixes) should be recognized as having a notrack prefix in 64-bit mode but not in 32-bit mode. 

Intel's documentation doesn't make it clear whether the notrack prefix still retains its old segment-override function when CET is enabled in legacy/compatibility mode. A couple of examples of 32-bit instruction encodings where that distinction would matter are:
* 3E FF 55 00 : is "notrack call ds:[ebp]" if it is recognized as a segment prefix or "notrack call ss:[ebp]" if it isn't.
* 26 3E FF 10 : is "notrack call ds:[eax]" if it is recognized as a segment prefix or "notrack call es:[eax]" if it isn't.

flobernd:
Fixed with #204

PR #204, opened 2021-08-28
Add support for `CET` `no-track` prefix

flobernd:
see #200


src/Decoder.c
```diff
@@ -2313,13 +2313,24 @@ static void ZydisSetAttributes(ZydisDecoderContext* context, ZydisDecodedInstruc
...
             }
-        } else
+        }
+
+        if (def->accepts_NOTRACK)
         {
-            if (def->accepts_segment)
+            instruction->attributes |= ZYDIS_ATTRIB_ACCEPTS_NOTRACK;
+            if (context->decoder->decoder_mode[ZYDIS_DECODER_MODE_CET] &&
+                (context->prefixes.effective_segment == 0x3E))
```
flobernd, line 2322: Note: `effecctive_segment` always prioritizes `0x64`/`0x65` in 64-bit mode which is perfect for our case regarding Intel SDM:

> In 64-bit mode, the 3EH prefix on an indirect call or jmp is recognized as a no-track prefix when the following 
conditions are satisfied. 
> -  3EH must be the last legacy prefix of any group (except any REX). 
> -  There must not be a 64H/65H prefix on the instruction. 
>
> In legacy/compatibility mode, the 3EH prefix on an indirect call or jmp is recognized as a no-track prefix 
when it is the last group 2 prefix on the instruction.

flobernd:
I probably missed some of the more uncommon indirect branch instructions:
https://github.com/zyantific/zydis-db/commit/cf5f8583f1a84a829a8369f91e4e5a7d83145645

Not sure about the `seg:offset` (`PTR`) based `jmp`/`call` as well.

src/Decoder.c
```diff
@@ -2313,13 +2313,24 @@ static void ZydisSetAttributes(ZydisDecoderContext* context, ZydisDecodedInstruc
...
             {
-                instruction->attributes |= ZYDIS_ATTRIB_ACCEPTS_SEGMENT;
+                instruction->attributes |= ZYDIS_ATTRIB_HAS_NOTRACK;
+                instruction->raw.prefixes[context->prefixes.offset_group2].type =
+                    ZYDIS_PREFIX_TYPE_EFFECTIVE;
             }
-            if (context->prefixes.effective_segment && def->accepts_segment)
+        }
+
+        if (def->accepts_segment && !def->accepts_branch_hints)
```
flobernd, line 2330: Regarding @tremalrik comment:

> Intel's documentation doesn't make it clear whether the notrack prefix still retains its old segment-override function when CET is enabled in legacy/compatibility mode. A couple of examples of 32-bit instruction encodings where that distinction would matter are:
>
> - 3E FF 55 00 : is "notrack call ds:[ebp]" if it is recognized as a segment prefix or "notrack call ss:[ebp]" if it isn't.
> - 26 3E FF 10 : is "notrack call ds:[eax]" if it is recognized as a segment prefix or "notrack call es:[eax]" if it isn't.

We could add another condition (`&& !def->accepts_NOTRACK`) here to force the 2nd case. For now, the segment override will be honored by Zydis until I find more information regarding this case.

src/Decoder.c
```diff
@@ -2313,13 +2313,25 @@ static void ZydisSetAttributes(ZydisDecoderContext* context, ZydisDecodedInstruc
...
+        {
+            instruction->attributes |= ZYDIS_ATTRIB_ACCEPTS_SEGMENT;
+            if (context->prefixes.effective_segment && 
+                !(instruction->attributes & ZYDIS_ATTRIB_HAS_NOTRACK))
```
flobernd, line 2334: Regarding @tremalrik comment:

> Intel's documentation doesn't make it clear whether the notrack prefix still retains its old segment-override function when CET is enabled in legacy/compatibility mode. A couple of examples of 32-bit instruction encodings where that distinction would matter are:
>
> - 3E FF 55 00 : is "notrack call ds:[ebp]" if it is recognized as a segment prefix or "notrack call ss:[ebp]" if it isn't.
> - 26 3E FF 10 : is "notrack call ds:[eax]" if it is recognized as a segment prefix or "notrack call es:[eax]" if it isn't.

For me it makes most sense, if the default group 2 behavior applies to the `no-track` prefix as well: Only last group 2 prefix counts, rest is ignored (special case in 64-bit mode, where `0x64`/`0x65` always has precedence). I did not test this on actual hardware and did not find any information either tho.

Zydis will output:
- `3E FF 55 00` : `notrack call dword ptr ss:[ebp]`
- `26 3E FF 10` : `notrack call dword ptr ds:[eax]`

athre0z:
With the current code, in 64-bit mode, `3e ff 10` disassembles to `notrack call qword ptr ds:[rax]` (which is expected, but `3e 2e ff 10` is decoded to `call qword ptr cs:[rax]`. According the the SDM:

> In 64-bit mode, the 3EH prefix on an indirect CALL or JMP is recognized as a no-track prefix if there isn’t a 64H/65H prefix on the instruction.

is seems to me like the second sample should have `notrack` as well, since 2e is neither 64 nor 65.

flobernd:
@athre0z 
> is seems to me like the second sample should have notrack as well, since 2e is neither 64 nor 65

Right ... another exception to the normal group 2 prefix handling here ... sight.

athre0z:
> I probably missed some of the more uncommon indirect branch instructions:
> [zyantific/zydis-db@cf5f858](https://github.com/zyantific/zydis-db/commit/cf5f8583f1a84a829a8369f91e4e5a7d83145645)
> 
> Not sure about the `seg:offset` (`PTR`) based `jmp`/`call` as well.

Screen Shot 2021-08-29 at 20 00 58

Looks exhaustive!

athre0z (approved):
LGTM now, thanks! :) I added a few regression tests a few important critical cases.

PR #207, opened 2021-08-31
Print asterisk in front of absolute `jmp`/`call` address (AT&T)

### LINKED ISSUE #176: Format the asterisk for AT&T absolute jumps/calls
Is it possible to format the asterisk for absolute jumps/calls in AT&T syntax?

For example:
 
    jmp *0x12345(,%rbp,8)    # With asterisk
    jmp 0x12345(,%rbp,8)     # Without asterisk (ZyDis default formatting)

Without the asterisk, the resulting disassembly can be ambiguous.

For example:

    E940230100               # jmp 0x12345     (relative jump with offset 0x12345)
    FF242545230100           # jmp *0x12345    (jump to the address stored at the absolute address 0x12345)

Currently, ZyDis formats both instructions as `jmp 0x12345`.

flobernd:
Fixes #176


flobernd:
I'm not quite sure if this syntax should apply to conditional branches as well or only to `jmp`/`call`.

athre0z (approved):
LGTM as well, thanks!

PR #209, opened 2021-09-01
`XOP`/`VEX`/`EVEX` is invalid in 16-bit real mode

### LINKED ISSUE #208: VEX/EVEX instruction decoding allowed in 16-bit real mode
Presence of VEX/EVEX prefix guarantees #UD in real mode but Zydis doesn't enforce that when initialized with `ZYDIS_MACHINE_MODE_REAL_16`. I believe that `DECODING_ERROR` should be returned for those instructions.

For some reason Intel provided very clearly labeled sections about some interactions (`VEX and the LOCK prefix`, `VEX and the 66H, F2H, and F3H prefixes`, `VEX and the REX prefix`) but didn't describe real mode behavior explicitly. All VEX/EVEX have exception classes assigned though and all of them guarantee #UD due to VEX/EVEX prefix.

This could be fixed in code or in the generator (treat all VEX/EVEX instrictions as if they had `protected_mode` flag). It seems that fix inside `ZydisNodeHandlerOpcode` would be the best approach allowing to fail as soon as possible.

flobernd:
Fixes #208

### DIFF
--- a/src/Decoder.c
+++ b/src/Decoder.c
@@ -457,6 +457,12 @@ static ZyanStatus ZydisDecodeXOP(ZydisDecoderContext* context,
     ZYAN_ASSERT(((data[1] >> 0) & 0x1F) >= 8);
     ZYAN_ASSERT(instruction->raw.xop.offset == instruction->length - 3);
 
+    if (context->decoder->machine_mode == ZYDIS_MACHINE_MODE_REAL_16)
+    {
+        // XOP is invalid in 16-bit real mode
+        return ZYDIS_STATUS_DECODING_ERROR;
+    }
+
     instruction->attributes |= ZYDIS_ATTRIB_HAS_XOP;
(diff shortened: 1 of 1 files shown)


athre0z (commented):
LGTM in general. Do we want the check for MVEX as well? Scrolling through the documentation, pretty much every single MVEX instruction seems to be documented with:

Screen Shot 2021-09-02 at 08 32 38

flobernd:
@athre0z EVEX and MVEX is only valid in 64-bit mode. We have a check for this somewhere in the main decoder loop. Thinking of that, the check I added for EVEX in this PR might even be redundant 😋

athre0z:
Can't confirm that: MVEX in 16 bit mode is being decoded:

Screen Shot 2021-09-02 at 08 53 00

flobernd:
Yes, I just checked as well and I will remove the check from the EVEX decoding as well. Same story here, will just decode as `bound` for non 64-bit mode.

PR #214, opened 2021-09-04
Add `AVX-512 FP16` isa-set

flobernd:


athre0z (approved):
LGTM! I didn't look through the individual definitions, however.

Issue #215, opened 2021-09-04
Assertion failure on new AVX512-FP16 VGETMANTPH instruction


tremalrik:
Running `ZydisInfo -64  62 B3 7C 39 26 00 00` using the new 61731ae version of Zydis, I get an assertion:
```
ZydisInfo: /home/tremalrik/x86_decode/zydis-master/zydis/src/Decoder.c:2567: ZydisSetAVXInformation: Assertion `!evex_b || ((!evex_w && context->evex.element_size == 32) || ( evex_w && context->evex.element_size == 64))' failed.
Aborted (core dumped)
```
XED decodes this instruction encoding to `vgetmantph ymm0{k1}, word ptr [rax]{1to16}, 0x0`.

flobernd:
Nice finding again! Almost expected an assert to fail after the latest change, but probably I did not test and fuzz long enough.

May I ask if you found this "randomly" or do you have some kind of testsuite?

tremalrik:
Yeah, I do have a directed-random fuzz tester (biased in favor of creating x86 instruction encodings with many prefixes and/or one of the multi-byte prefixes - 0F_0F, 0F_38, 0F_3A, VEX2, VEX3, XOP, EVEX - since these parts of the x86 opcode space tend to be unreasonably hard to hit with unbiased-random byte sequences). Adapting it to specifically test Zydis itself turned out to be fairly simple, see https://github.com/tremalrik/zydis-fuzzer for an initial version. For Zydis, this fuzzer seems only really useful for finding decode assertions - it will not detect incorrect decodings or anything of that sort.

With the 61731ae version of Zydis, it generally finds an assertion in about 1 to 15 seconds, depending on input seed. So far, I've seen it produce assertions on a VRNDSCALEPH (`62 53 7C 3C 08 47 75 11` in 64-bit mode), a VREDUCEPH (`36 67 62 E3 7C DF 56 9E 8E 79 C2 46 A6` in 16-bit mode) and a VCMPPH (`62 C3 14 19 C2 4B 2F 32` in 16-bit mode) in addiiton to the VGETMANTPH.

flobernd:
That looks very useful :-)

@athre0z Maybe we should think about adding this custom fuzzer to our CI as well, if @tremalrik does not mind.

OssFuzz indeed seems not reach the more complex regions of the X86 ISA within an acceptable amount of time.

athre0z:
[...]

PR #217, opened 2021-09-05
Add `ZYDIS_FORCE_ASSERTS` CMake option

athre0z:
This option forces assertions to be enabled even in release builds. It's a bit of a hack because it touches the global `{C,CXX}FLAGS`, but since it's only intended for use in our fuzzing pipeline, this should be fine.

Related discussion in #215.


flobernd (approved):
LGTM. 

Maybe it makes sense to move this option to Zycore and use it for all projects.

Should not make a difference for Zydis, but as this option basically affects to whole build, it should be on the core dependency.

PR #218, opened 2021-09-05
CI: test build of VS projects in `msvc` directory

athre0z:
This PR contains two commits.

The first is general clean-up of the CI configuration and updates all images to their latest version. I switched all images that were previously on rolling images (`xxx-latest`) to a hard-coded version. An exception to this is the macOS image, where I remember reading that using `-latest` is strongly recommended because of GH gets rid of old macOS versions very quickly (because they have to run it on Mac Minis instead of virtualized boxes).

The second commit is the actual core of this PR: it adds build targets for the VS projects in the `msvc` directory, also building the kernel mode driver.

I'm currently trying to figure out how to motivate `msbuid` to build more than just the kernel mode driver, which is proving difficult by just brute-forcing via the force-pushing to the CI. Setting up a Windows VM while writing this.

CC @Mattiwatti

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,45 +3,44 @@ name: GitHub Actions CI
 on: [push, pull_request]
 
 jobs:
-  build:
-    name: "Build Zydis (${{ matrix.image_name }}, ${{ matrix.no_libc }})"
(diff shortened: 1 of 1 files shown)


flobernd (approved):
Nice! LGTM.

PR #227, opened 2021-09-09
Fixed MVEX rounding mode decoding (Fixes #226)

### LINKED ISSUE #226: Out of range rounding mode values may be reported by decoder
Zydis doesn't mask out highest bit (SAE bit) of `mvex.sss` when decoding rounding mode. This results in invalid enum value being reported in decoded instruction.

Reproduce with: `ZydisInfo.exe -64 62F271E950C2`. ZydisInfo will crash due to insufficient error checking.

mappzor:
Fixes #226


src/Decoder.c
```diff
@@ -3082,8 +3082,8 @@ static void ZydisSetAVXInformation(ZydisDecoderContext* context,
             // Nothing to do here
             break;
         case ZYDIS_MVEX_FUNC_RC:
-            instruction->avx.rounding.mode = ZYDIS_ROUNDING_MODE_RN + instruction->raw.mvex.SSS;
-            break;
+            instruction->avx.rounding.mode = ZYDIS_ROUNDING_MODE_RN + (instruction->raw.mvex.SSS & 3);
+            // Intentional fallthrough
```
flobernd, line 3086: Fyi: We have a ZYAN_FALLTHROUGH macro for this purpose 😃

If you don't want to edit your PR, I can as well change that line after merging.

mappzor: I knew I've seen this macro somewhere. Actually it turns out it's used exactly everywhere except the place I used for reference when making this PR :grinning: Will update both.

athre0z (approved):
LGTM, thanks! :)

PR #229, opened 2021-09-10
Fixed issues with enum signedness

mappzor:
Enhanced some checks that were ineffective when compiled under MSVC which has signed enums.

I also added a check for maximum padding size because padding values are later casted to ZyanU8. ZyCore doesn't provide min/max defines for integer types, so to avoid forced dependence on stdint I simply used hardcoded value. Let me know if you would like to resolve that differently.

### DIFF
--- a/src/Decoder.c
+++ b/src/Decoder.c
@@ -5063,7 +5063,7 @@ ZyanStatus ZydisDecoderInit(ZydisDecoder* decoder, ZydisMachineMode machine_mode
 
 ZyanStatus ZydisDecoderEnableMode(ZydisDecoder* decoder, ZydisDecoderMode mode, ZyanBool enabled)
 {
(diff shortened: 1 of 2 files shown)


src/Decoder.c
```diff
@@ -5063,7 +5063,7 @@ ZyanStatus ZydisDecoderInit(ZydisDecoder* decoder, ZydisMachineMode machine_mode
 
 ZyanStatus ZydisDecoderEnableMode(ZydisDecoder* decoder, ZydisDecoderMode mode, ZyanBool enabled)
 {
-    if (!decoder || (mode < 0) || (mode > ZYDIS_DECODER_MODE_MAX_VALUE))
+    if (!decoder || ((ZyanUSize)mode > ZYDIS_DECODER_MODE_MAX_VALUE))
```
mappzor, line 5066: This is just for consistency.

athre0z (approved):
Thanks! :)

PR #233, opened 2021-09-15
Fixed {sae}/{rc} formatting (Fixes #232)

### LINKED ISSUE #232: EVEX-encoded VCOMISS and the {sae} modifier
According to the Intel SDM (June 2021 revision) description of the `vcomiss` instruction, the EVEX-encoded version of this instruction supports the `{sae}` (suppress-all-exceptions) modifier when used with two register arguments. This is recognized by xed but not Zydis:
 * `xed -64 -d 62 f1 7c 78 2f c1` returns `vcomiss xmm0{sae}, xmm1`
 * `ZydisInfo -64 62 f1 7c 78 2f c1` returns `vcomiss xmm0, xmm1`

The same issue also applies to the closely related instructions `vcomish`, `vcomisd`, `vucomiss`, `vucomish`, `vucomisd`.

(The SDM appears to contradict itself on this modifier: it specifies that these instructions belong to the "E3NF" exception class, but the description of E3NF (volume 2, section 2.7.3, table 2-47) specifies that EVEX.b != 0 should result in #UD. This would seem to prohibit the `{sae}` encoding completely, but this contradicts both the instruction description and the xed output, so I assume the table 2-47 is in error.)

mappzor:
Fixes #232


flobernd:
LGTM! Thans a lot for the fix 😃

PR #234, opened 2021-09-16
Do not sign ZydisWinKernel sample

Mattiwatti:
This PR is an attempt to fix the CI build by disabling driver signing at the MSVC/MSBuild project level, as opposed to the `.vcxproj.user` files MSVC normally stores user settings in when the defaults are overridden.


athre0z:
Seems to do the job -- perfect! Thanks Matti! :)

PR #238, opened 2021-09-19
Fix general address-width handling

### LINKED ISSUE #237: Decoder can be initialized with ill-formed defaults
`ZydisDecoderInit` accepts `machine_mode` and `address_width`. In conjunction those arguments will determine decoder's default values for stack size (ssz), operand size (osz) and address size (asz). Typically Zydis users pass matching values (e.g. 32-bit machine mode with 32-bit address width) but decoder logic acts improperly when those values do not match:
(decoder->machine_mode, decoder->address_width) -> (ssz, osz, asz)
(16, 32) -> (32, 16, 32)
(32, 16) -> (16, 32, 16)

Such defaults are not possible on real hardware. They are controlled by D/B flag in CS and SS. For CS we have CS.D that controls "effective addresses and operands referenced by instructions in the segment", so default values of osz and asz, should be always the same. For SS we have SS.B that determines ssz.

Expected behavior would be to derive osz and asz from `machine_mode` and take ssz from `address_width` (which probably should be named stack_width just as in decoded instruction).

flobernd:
- Default address-width is now determined by the machine-mode
- Default stack size can be passed `ZydisDecoderInit` instead

Closes #237


flobernd:
@athre0z Please don't merge it before the next major release as it contains a breaking API change.

PR #242, opened 2021-09-29
Fixed no_source_dest_match check (Fixes #241)

### LINKED ISSUE #241: no_source_dest_match doesn't consider memory operands
Validation logic for second source operand should be executed for reg forms only. Currently it malfunctions when memory operand is present e.g.
```
vfmaddcph zmm15 {k6}, zmm21, dword ptr ds:[r15+0x19C] {1to16}
-64 62565656567C6767 -> SIB form decodes fine (because only modrm is checked)
-64 6256565656BF9C010000 ->ZYDIS_STATUS_BAD_REGISTER
```

mappzor:
Fixes #241


flobernd:
Hey there, thanks for the PR! I think it might be better to fix this in the table generator instead. Seems like the flag is incorrectly set for the non-reg version. Let me check this tomorrow, before I merge :-)

mappzor:
I thought about this but came to the conclusion that flag must be set for those versions because dest and first source operand are still registers, so they need validation.

flobernd:
Indeed. Thought about combining some of these validation flags to an enum (distinct regs reg, distinct regs mem, vsib).

Merging for now :-) The vsib flag is not on the legacy struct anyways, so some shuffling would be required.

flobernd:
Merged. Thanks again!

PR #248, opened 2021-10-08
build(cmake): fix PUBLIC include dir of installed lib

Tachi107:
See https://github.com/zyantific/zydis/pull/191#discussion_r724288380 for details


athre0z:
Thanks again, merged! :)

PR #249, opened 2021-10-08
build(cmake): add version and soversion to the library

Tachi107:
This adds versioning information to the built library.

When building the shared library on Linux systems, two new objects will be created: `libZydis.so.3.1.0.0` and `libZydis.so.3.1`.

This is really useful when talking about ABI compatibility.

The variables `Zydis_VERSION`, `Zydis_VERSION_MAJOR` and `Zydis_VERSION_MINOR` are implicitly created when calling `project(Zydis VERSION x.y.z)`

Similar to https://github.com/zyantific/zycore-c/pull/31


athre0z:
Thanks! Could you update the zycore submodule to also include your PR over there that I just merged?

PR #259, opened 2021-10-29
Fix build on macOS 12

athre0z:
This PR disables struct packing for `ZydisShortString` on Apple aarch64 platforms. For an unknown reason, compilation is now failing with `ld` complaining about pointer alignment starting with macOS 12 on aarch64. On macOS 11 + aarch64, things compile perfectly fine without the change.

While we have many other packed structs, the linker seems to particularly take offense with this one, as it is the only packed struct containing a pointer. I think it might be related to some form of pointer authentication.

The binary bloat from this appears to be about 40KiB.

### DIFF
--- a/include/Zydis/ShortString.h
+++ b/include/Zydis/ShortString.h
@@ -45,7 +45,9 @@ extern "C" {
 /* Enums and types                                                                                */
 /* ============================================================================================== */
 
-#pragma pack(push, 1)
+#if !(defined(ZYAN_AARCH64) && defined(ZYAN_APPLE))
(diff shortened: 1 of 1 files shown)


flobernd (approved):
Maybe as an improvement, we could prepend the length to the actual strings (this should be easily possible as we generate the strings anyways) and remove the 'length' field from the struct. In this case a struct would not even be necessary, but probably still desired as a hint.

msuiche:
Are you guys planning to release a new crate too? https://crates.io/crates/zydis/versions

PR #267, opened 2021-11-12
docs: add man pages

Tachi107:
Zydis ships two command line tools, but doesn't provide any man page. Since Debian enforces that every binary must also provide a man page here I am providing a patch.

Standard UNIX manpages are painful to write, but luckly there are tools like Ronn(-NG) or Pandoc that allow to write the page in simple Markdown and convert it in roff, the format that `man` uses.

These man pages are extremely basic, since I'm not very familiar with the tools and how they work :/
Please don't hesitate making suggestions :)

Since Ronn-NG is an extra dependency the manpages are not built by default, but only if requested with `ZYDIS_BUILD_MAN`.

Ronn-NG is required instead of the old Ronn because the latter always generates output files in the same dir as the input files, and that doesn't play nice with CMake and out of tree builds, while the -NG version has a `--output-dir` option.

### DIFF
--- a/man/ZydisInfo.1.ronn
+++ b/man/ZydisInfo.1.ronn
@@ -1,0 +1,57 @@
+ZydisInfo(1) -- detailed instruction information
+================================================
+
+## SYNOPSIS
+
+`ZydisInfo` <machine_mode> [<stack_width>] <hexbytes>
+
+## DESCRIPTION
+
+`ZydisInfo` allows you to decode X86 & X86-64 assembly displaying lots of information about it.
+
+## OPTIONS
+
+`ZydisInfo` supports four different machine modes
+
+  * `-real`:
+    real machine mode
+
+  * `-16`:
+    16 bits machine mode
+
+  * `-32`:
+    32 bits machine mode
+
+  * `-64`:
+    64 bits machine mode
+
+You can also specify the stack width one of the following options
+
+  * `-16`:
+    16 bits
+
+  * `-32`:
+    32 bits
+  
+  * `-64`:
+    64 bits
+
+## EXAMPLES
+
+    $ ZydisInfo -64 66 3E 65 2E F0 F2 F3 48 01 A4 98 2C 01 00 00
+    == [    BASIC ] ==================================================
+       MNEMONIC: add [ENC: DEFAULT, MAP: DEFAULT, OPC: 0x01]
+         LENGTH: 15
+            SSZ: 64
+           EOSZ: 64
+           EASZ: 64
+       CATEGORY: BINARY
+        ISA-SET: I86
+        ISA-EXT: BASE
+     EXCEPTIONS: NONE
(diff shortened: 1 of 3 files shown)


athre0z (commented):
Thanks for the PR! LGTM in general, I left a few minor comments.

Should the files have `.ronn` extension? Since they seem to be vanilla Markdown, if that works, I think it would make sense to have them end in `.md` to make GitHub web render them if you view them.

> Standard UNIX manpages are painful to write, but luckly there are tools like Ronn(-NG) or Pandoc that allow to write the page in simple Markdown and convert it in roff, the format that man uses.

I just looked up an example and am inclined to agree -- it looks pretty terrible indeed. Good choice to go with a transpiler approach!

> These man pages are extremely basic, since I'm not very familiar with the tools and how they work :/

Neither are we, so I guess we'll just merge things and wait for someone with more experience to flame us for all that we did wrong! :)

man/ZydisInfo.1.ronn
```diff
@@ -0,0 +1,30 @@
+ZydisInfo(1) -- disassemble in your terminal
+============================================
+
+## SYNOPSIS
+
+`ZydisInfo -[real|16|32|64] <hexbytes>`
```
mappzor, line 6: Current syntax is:
```
ZydisInfo <machine_mode> [stack_width] <hexbytes>
Machine mode:      -real|-16|-32|-64
Stack width:       -16|-32|-64
```

Tachi107:
> Should the files have `.ronn` extension? Since they seem to be vanilla Markdown, if that works, I think it would make sense to have them end in `.md` to make GitHub web render them if you view them.

Yes, because they aren't actually pure Markdown, it has a few manpage-specific features (like special rendering of `<text>`). Luckily GitHub renders them anyway 

> > These man pages are extremely basic, since I'm not very familiar with the tools and how they work :/
> 
> Neither are we, so I guess we'll just merge things and wait for someone with more experience to flame us for all that we did wrong! :)

Haha, good idea

I'll address your feedback as soon as possible

man/ZydisInfo.1.ronn
Tachi107: Do these options also apply to ZydisDisasm?

Tachi107:
Tried to apply your suggestions :)

Please note that this is the first time I "write" manpages, and I'm also not familiar with the terminology used by Zydis (I've never done anything serious with assembly language, unfortunately), so any corrections will always be appreciated (even grammatical ones, I'm not a native English speaker)

man/ZydisInfo.1.ronn
flobernd: For ZydisDisasm I never implemented the stack-width option if I remember correctly.

Might be a nice addition tho. Will keep that in mind for future improvement.

athre0z (approved):
Thanks, LGTM!

PR #275, opened 2021-11-21
Add regression tests for recent bugs

athre0z:


flobernd (approved):
LGTM!

We should probably improve the regression tests to somehow have error cases as well.

In my last PR this would have been helpful to spot the invalid register that was accepted by the decoder anyways.

athre0z:
Yeah, sounds sensible and shouldn't be too much work. Should probably create a separate issue for it so we don't forget.

mappzor:
I will add a negative case for TMM stuff in a separate PR (need to add that constraint to encoder anyway).

PR #277, opened 2021-11-26
add WebAssembly support

williballenthin:
Relying on the changes in https://github.com/zyantific/zycore-c/pull/40, this PR adds support to Zydis for building for WebAssembly. The only updates are to implement the appropriate `ZydisStringAppend*` routines when wasm is detected. Because wasm is most like a 32-bit architecture, we go the 32-bit-optimized `U32` route.

### DIFF
--- a/src/String.c
+++ b/src/String.c
@@ -65,7 +65,7 @@ static const char* const DECIMAL_LOOKUP =
 /* Decimal                                                                                        */
 /* ---------------------------------------------------------------------------------------------- */
 
-#if defined(ZYAN_X86) || defined(ZYAN_ARM) || defined(ZYAN_EMSCRIPTEN)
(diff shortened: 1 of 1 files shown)


athre0z (approved):
LGTM, thanks again! I updated the submodule and rebased on master so we can run tests properly. Your branch was forked off a master that was 179 commits behind, presumably accidentally from an old fork.

williballenthin:
> Your branch was forked off a master that was 179 commits behind,

For context, I was working from the snapshot used by zydis-rs. While GH showed that the PR applied cleanly, I wasn’t sure if there were any new code/changes that need to be updated too. 

Thanks for the reviews!

athre0z:
Ah, makes sense. Thanks for the clarification!

PR #278, opened 2021-12-01
Decouple operand decoding

### LINKED ISSUE #110: Structual changes to decouple operand-decoding
Decouple operand-decoding from general instruction decoding to significantly decrease the size of `ZydisDecodedInstruction` which will improve usage of CPU cache in some situations.

---

- [x] Remove the `ZydisDecodedOperand` array from the `ZydisDecodedInstruction` struct
- [x] Provide a function that allows to decode `count` operands beginning from a given `index`. Additional flags like e.g. "include hidden operands" could be useful as well.
- [x] Implement a wrapper struct that contains the fixed size `ZydisDecodedOperand` array (like in the current `ZydisDecodedInstruction`) and a function to fill this struct (for convenience reasons)

---

Possible prototypes for the decoding functions could be:

```c
ZydisStatus ZydisDecodeOperands(ZydisDecodedInstruction* instruction, ZydisDecodedOperand* operands, ZyanU8 count);
```

and

```c
ZydisStatus ZydisDecodeOperandsEx(ZydisDecodedInstruction* instruction, ZydisDecodedOperand* operands, ZyanU8 count, ZyanU8 index, ZydisOperandDecodeFlags flags);
```

where `ZydisOperandDecodeFlags` might look like this:

```c
#define ZYDIS_OPERAND_DECODE_FLAGS_VISIBLE 1 << 0
#define ZYDIS_OPERAND_DECODE_FLAGS_HIDDEN  1 << 1
#define ZYDIS_OPERAND_DECODE_FLAGS_ALL     = ZYDIS_OPERAND_DECODE_FLAGS_VISIBLE | ZYDIS_OPERAND_DECODE_FLAGS_HIDDEN 
```

Additionally the basic `ZydisDecodedInstruction` struct should always be filled with at least `operand_count` and `operand_count_hidden`.

flobernd:
Closes #110

### DIFF
--- a/include/Zydis/Decoder.h
+++ b/include/Zydis/Decoder.h
@@ -148,6 +148,20 @@ typedef enum ZydisDecoderMode_
 } ZydisDecoderMode;
 
 /* ---------------------------------------------------------------------------------------------- */
+/* Decoder flags                                                                                  */
+/* ---------------------------------------------------------------------------------------------- */
+
+/**
+ * Defines the `ZydisDecodingFlags` data-type.
+ */
+typedef ZyanU8 ZydisDecodingFlags;
+
+/**
+ * Only decode visible operands.
+ */
+#define ZYDIS_DFLAG_VISIBLE_OPERANDS_ONLY   (1 << 0)
+
+/* ---------------------------------------------------------------------------------------------- */
 /* Decoder struct                                                                                 */
 /* ---------------------------------------------------------------------------------------------- */
 
@@ -210,9 +224,54 @@ ZYDIS_EXPORT ZyanStatus ZydisDecoderEnableMode(ZydisDecoder* decoder, ZydisDecod
     ZyanBool enabled);
 
 /**
+ * Decodes the instruction in the given input `buffer` and returns all details (e.g. operands).
+ *
+ * @param   decoder         A pointer to the `ZydisDecoder` instance.
+ * @param   buffer          A pointer to the input buffer.
+ * @param   length          The length of the input buffer. Note that this can be bigger than the
+ *                          actual size of the instruction -- you don't have to know the size up
+ *                          front. This length is merely used to prevent Zydis from doing
+ *                          out-of-bounds reads on your buffer.
+ * @param   instruction     A pointer to the `ZydisDecodedInstruction` struct, that receives the
+ *                          details about the decoded instruction.
+ * @param   operands        The array that receives the decoded operands.
+ *                          Refer to `ZYDIS_MAX_OPERAND_COUNT` or `ZYDIS_MAX_OPERAND_COUNT_VISIBLE`
+ *                          when allocating space for the array to ensure that the buffer size is
+ *                          sufficient to always fit all instruction operands.
+ * @param   operand_count   The length of the `operands` array.
+ *                          This argument as well limits the maximum amount of operands to decode.
+ *                          If this value is `0`, no operands will be decoded and `ZYAN_NULL` will
+ *                          be accepted for the `operands` argument.
+ * @param   flags           Additional decoding flags.
+ *
+ * This function decodes `MIN(operand_count, instruction.operand_count)` operands. The excess
+ * items in the `operands` array are not zeroed. The `instruction.operand_count` field must be
+ * checked in addition to the passed `operand_count`, to determine the actual amount of decoded
+ * operands.
+ *
+ * The `ZYDIS_DFLAG_VISIBLE_OPERANDS_ONLY` can be passed to only decode visible operands. In this
+ * case `MIN(operand_count, instruction.operand_count_visible)` operands are decoded by this
+ * function and the `instruction.operand_count_visible` field must be checked in addition to the
+ * passed `operand_count`, to determine the actual amount of decoded operands.
+ *
+ * Please refer to `ZydisDecoderDecodeInstruction` and `ZydisDecoderDecodeOperands`, if operand
+ * decoding is not required or should be done separately.
+ *
+ * This function is not available in MINIMAL_MODE.
+ *
+ * @return  A zyan status code.
+ */
+ZYDIS_EXPORT ZyanStatus ZydisDecoderDecodeFull(const ZydisDecoder* decoder,
+    const void* buffer, ZyanUSize length, ZydisDecodedInstruction* instruction,
+    ZydisDecodedOperand* operands, ZyanU8 operand_count, ZydisDecodingFlags flags);
+
+/**
  * Decodes the instruction in the given input `buffer`.
  *
  * @param   decoder     A pointer to the `ZydisDecoder` instance.
+ * @param   context     A pointer to a decoder context struct which is required for further
+ *                      decoding (e.g. operand decoding using `ZydisDecoderDecodeOperands`) or
+ *                      `ZYAN_NULL` if not needed.
  * @param   buffer      A pointer to the input buffer.
  * @param   length      The length of the input buffer. Note that this can be bigger than the
  *                      actual size of the instruction -- you don't have to know the size up
@@ -223,8 +282,39 @@ ZYDIS_EXPORT ZyanStatus ZydisDecoderEnableMode(ZydisDecoder* decoder, ZydisDecod
  *
  * @return  A zyan status code.
  */
-ZYDIS_EXPORT ZyanStatus ZydisDecoderDecodeBuffer(const ZydisDecoder* decoder,
-    const void* buffer, ZyanUSize length, ZydisDecodedInstruction* instruction);
+ZYDIS_EXPORT ZyanStatus ZydisDecoderDecodeInstruction(const ZydisDecoder* decoder,
+    ZydisDecoderContext* context, const void* buffer, ZyanUSize length,
+    ZydisDecodedInstruction* instruction);
+
+/**
+ * Decodes the instruction operands.
+ *
+ * @param   decoder         A pointer to the `ZydisDecoder` instance.
+ * @param   context         A pointer to the `ZydisDecoderContext` struct.
+ * @param   instruction     A pointer to the `ZydisDecodedInstruction` struct.
+ * @param   operands        The array that receives the decoded operands.
+ *                          Refer to `ZYDIS_MAX_OPERAND_COUNT` or `ZYDIS_MAX_OPERAND_COUNT_VISIBLE`
+ *                          when allocating space for the array to ensure that the buffer size is
+ *                          sufficient to always fit all instruction operands.
+ *                          Refer to `instruction.operand_count` or
+ *                          `instruction.operand_count_visible' when allocating space for the array
+ *                          to ensure that the buffer size is sufficient to fit all operands of
+ *                          the given instruction.
+ * @param   operand_count   The length of the `operands` array.
+ *                          This argument as well limits the maximum amount of operands to decode.
+ *                          If this value is `0`, no operands will be decoded and `ZYAN_NULL` will
+ *                          be accepted for the `operands` argument.
+ *
+ * This function fails, if `operand_count` is larger than the total number of operands for the
+ * given instruction (`instruction.operand_count`).
+ *
+ * This function is not available in MINIMAL_MODE.
+ *
+ * @return  A zyan status code.
+ */
+ZYDIS_EXPORT ZyanStatus ZydisDecoderDecodeOperands(const ZydisDecoder* decoder,
+    const ZydisDecoderContext* context, const ZydisDecodedInstruction* instruction,
+    ZydisDecodedOperand* operands, ZyanU8 operand_count);
 
 /** @} */
 
--- a/include/Zydis/DecoderTypes.h
+++ b/include/Zydis/DecoderTypes.h
@@ -720,16 +720,19 @@ typedef struct ZydisDecodedInstruction_
     ZyanU8 address_width;
     /**
      * The number of instruction-operands.
+     *
+     * Explicit and implicit operands are guaranteed to be in the front and ordered as they are
+     * printed by the formatter in `Intel` mode. No assumptions can be made about the order of
+     * hidden operands, except that they always located behind the explicit and implicit operands.
      */
     ZyanU8 operand_count;
     /**
-     * Detailed info for all instruction operands.
+     * The number of explicit (visible) instruction-operands.
      *
-     * Explicit operands are guaranteed to be in the front and ordered as they are printed
-     * by the formatter in Intel mode. No assumptions can be made about the order of hidden
-     * operands, except that they always located behind the explicit operands.
+     * Explicit and implicit operands are guaranteed to be in the front and ordered as they are
+     * printed by the formatter in `Intel` mode.
      */
-    ZydisDecodedOperand operands[ZYDIS_MAX_OPERAND_COUNT];
+    ZyanU8 operand_count_visible;
     /**
      * Instruction attributes.
      */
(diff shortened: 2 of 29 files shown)


include/Zydis/DecoderTypes.h
```diff
@@ -720,16 +720,19 @@ typedef struct ZydisDecodedInstruction_
     ZyanU8 address_width;
     /**
      * The number of instruction-operands.
+     *
+     * Explicit- and implicit operands are guaranteed to be in the front and ordered as they are
+     * printed by the formatter in `Intel` mode. No assumptions can be made about the order of
+     * hidden operands, except that they always located behind the explicit operands.
```
athre0z, line 726:
```suggestion
     * Explicit and implicit operands are guaranteed to be in the front and ordered as they are
     * printed by the formatter in `Intel` mode. No assumptions can be made about the order of
     * hidden operands, except that they always located behind the explicit and implicit operands.
```

include/Zydis/Formatter.h
```diff
@@ -1046,15 +1056,16 @@ ZYDIS_EXPORT ZyanStatus ZydisFormatterFormatInstruction(const ZydisFormatter* fo
...
+    ZyanU8 operand_count, char* buffer, ZyanUSize length, ZyanU64 runtime_address,
+    void* user_data);
 
 /**
  * Formats the given operand and writes it into the output buffer.
  *
  * @param   formatter       A pointer to the `ZydisFormatter` instance.
  * @param   instruction     A pointer to the `ZydisDecodedInstruction` struct.
- * @param   index           The index of the operand to format.
+ * @param   operand         A pointer to the `ZydisDecodedOperand` struct of the operand to format.
```
athre0z, line 1068: Might be easier to understand like this:

```suggestion
 * @param   operand         A pointer to the array of operands.
```

athre0z (commented):
Excellent PR, all good changes. Reviewed everything except for the fuzzing logic, which I want to take a look at when I'm less tired, presumably sometime on the weekend.

Do we intend to keep the API as-is, or do we turn the `ZydisDecoderDecodeBuffer` function into `ZydisDecoderDecodeBufferEx` later, providing a simpler variant that automatically decodes the operands as well, naming it `ZydisDecoderDecodeBuffer` later? I originally thought this would be the way to go, but looking at the formatter interface, we'd also have to support the struct filled by the simple variant in that one, resulting in an explosion of complexity that I'd strongly prefer to avoid.

I'm kind of leaning towards keeping the interface as proposed in this PR as our proper interface and instead providing an even simpler function that fills an even larger struct that also does formatting in one wash, the audience of which would primarily be beginner users. 

In any case, I'd be happy to keep as-is and discuss this in a separate issue later.

include/Zydis/SharedTypes.h
```diff
@@ -46,8 +46,9 @@ extern "C" {
 /* Constants                                                                                      */
 /* ---------------------------------------------------------------------------------------------- */
 
-#define ZYDIS_MAX_INSTRUCTION_LENGTH 15
-#define ZYDIS_MAX_OPERAND_COUNT      10
+#define ZYDIS_MAX_INSTRUCTION_LENGTH    15
+#define ZYDIS_MAX_OPERAND_COUNT         10 // TODO: Auto generate
+#define ZYDIS_MAX_OPERAND_COUNT_VISIBLE  5 // TODO: Auto generate
```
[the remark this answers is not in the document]
mappzor: Yes, that's why `ZYDIS_ENCODER_MAX_OPERANDS` is also 5. Actually there is a bunch of them:

```valignd, valignq, vcmppd, vcmpph, vcmpps, vcmpsd, vcmpsh, vcmpss, vdbpsadbw, vfixupimmpd, vfixupimmps, vfixupimmsd, vfixupimmss, vgetmantsd, vgetmantsh, vgetmantss, vgf2p8affineinvqb, vgf2p8affineqb, vinsertf32x4, vinsertf32x8, vinsertf64x2, vinsertf64x4, vinserti32x4, vinserti32x8, vinserti64x2, vinserti64x4, vpalignr, vpcmpb, vpcmpd, vpcmpq, vpcmpub, vpcmpud, vpcmpuq, vpcmpuw, vpcmpw, vpermil2pd, vpermil2ps, vpshldd, vpshldq, vpshldw, vpshrdd, vpshrdq, vpshrdw, vpternlogd, vpternlogq, vrangepd, vrangeps, vrangesd, vrangess, vreducesd, vreducesh, vreducess, vrndscalesd, vrndscalesh, vrndscaless, vshuff32x4, vshuff64x2, vshufi32x4, vshufi64x2, vshufpd, vshufps```

**Review by mappzor — COMMENTED** [review]
Nice changes :)

Some thoughs about API (@athre0z):
> Do we intend to keep the API as-is, or do we turn the ZydisDecoderDecodeBuffer function into ZydisDecoderDecodeBufferEx later, providing a simpler variant that automatically decodes the operands as well, naming it ZydisDecoderDecodeBuffer later? I originally thought this would be the way to go, but looking at the formatter interface, we'd also have to support the struct filled by the simple variant in that one, resulting in an explosion of complexity that I'd strongly prefer to avoid.
> 
> I'm kind of leaning towards keeping the interface as proposed in this PR as our proper interface and instead providing an even simpler function that fills an even larger struct that also does formatting in one wash, the audience of which would primarily be beginner users.

What about this:
- `ZydisDecoderDecodeBuffer` -> function that decodes instruction + operands (simple utility function)
- `ZydisDecoderDecodeInstruction` -> minimal decoding (renamed `ZydisDecoderDecodeBuffer` from this PR)
- `ZydisDecoderDecodeOperands` -> as is in this PR

In this concept signature for `ZydisDecoderDecodeBuffer` would change to accept operands, so it would be another breaking change since v3. This has some advantages though:
- problem you highlighted (with having one big struct) is avoided
- function still does what it did in v3 - decodes instruction and operands, just the output is split into new v4-fashion
- this is easy to explain in porting guide and keeps migration rather staigtforward. For most existing users documentation of new `ZydisDecoderDecodeBuffer` would introduce them to a new concept of decoding instruction and operands separately while immediately informing them it's optional to do so because they can keep using `ZydisDecoderDecodeBuffer` after making minimal adjustments to existing codebases.

File: include/Zydis/SharedTypes.h (hunk 3 above)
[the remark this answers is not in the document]
Comment (flobernd) [reply include/Zydis/SharedTypes.h:51]: Some of them are very nice. They split the 8-bit immediate into two 4 bit values where one part is used as a register id and the other part as the actual immediate. AMD XOP ... 😄

File: include/Zydis/DecoderTypes.h [hunk 4]
```diff
@@ -1231,6 +1234,111 @@ typedef struct ZydisDecodedInstruction_
...
+         *
+         * This value is only set, if a memory operand is encoded in `modrm.rm` and the `SIB` byte
+         * is present.
+         */
+        ZyanU8 id_index;
+    } reg_info;
+    /**
+     * Internal EVEX-specific information.
+     */
+    struct
```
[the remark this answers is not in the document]
flobernd: Yes, that was intentional. I thought about this for quite some time. As we rely on the user to allocate our structs, we have to make sure the input struct size matches our internal expectations. In a pure C/C++ environment, we could probably `ifdef`-ing these members away without any problem, but that's a different story for bindings. We would always have to check if the struct size is matching (e.g. going the Windows API way and introducing a `size` field in the `ZydisDecodedInstruction` that needs to be initialized by the user - extra complexity which I don't really like).

I think we might want to discuss this in detail. One good solution might be to introduce a `ZYDIS_STRICT_C_UNSAFE_FOR_BINDINGS` option. This would indicate that `Zydis` is used in a pure C/C++ project and always built on demand (-> we know the feature defines of headers and code files are the same). This would as well allow us to e.g. change the `ZydisDecodedOperand` struct to `unions` or to introduce bit-fields in the public API.

src/Encoder.c
```diff
@@ -4404,14 +4405,14 @@ ZYDIS_EXPORT ZyanStatus ZydisEncoderDecodedInstructionToEncoderRequest(
     }
     request->allowed_encodings = 1 << instruction->encoding;
 
-    if (instruction->operand_count > ZYAN_ARRAY_LENGTH(instruction->operands))
+    if (operand_count > instruction->operand_count)
```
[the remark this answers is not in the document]
flobernd: I partially agree this check could be dangerous, but at the same time I think it's the best we can come up with. This check was added to the formatter in the exact same manner as well.

Relying on the `instruction.operand_count_visible` is even more dangerous imho. I had the following scenario in my mind:

1. Decode instruction (3 visible operands)
2. Deccode operands. User chooses to only decode 2 operands, `operands` array has a length of 2)
3. Format/Encode -> BAM!

Even if it's kinda redundant and still not perfectly safe, I personally like to always pass the size/length of arbitrary length buffers in public APIs. This way you can directly see the buffer and the length side by side and it should be clear to every user that the provided buffer must be large enough to match the desired count/length.

We can add an additional check for e.g. `ZYDIS_OPERAND_COUNT_MAX`/`ZYDIS_ENCODER_MAX_OPERANDS` on these places, but even then it would access invalid memory in my example above.

tools/ZydisFuzzEncoder.c
```diff
@@ -75,17 +77,24 @@ void ZydisCompareRequestToInstruction(const ZydisEncoderRequest *request,
...
+        {
+            fputs("Failed to decode instruction operands\n", ZYAN_STDERR);
+            abort();
+        }
         insn = &knc_insn;
+        operands = knc_operands;
     }
 
     ZyanU8 visible_operand_count = 0;
     while ((visible_operand_count < insn->operand_count) &&
```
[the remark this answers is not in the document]
flobernd: A right. That reminds me to check the logic of all changes places again. Was trying to get the project compiling (mainly by replacing/adding stuff) and did not double check if all the changes make sense.

flobernd:
@mappzor 

> What about this: [...]

I like the idea of providing one function that combines instruction- and operand-decoding. However I would probably just name the utility function to something like `ZydisDecoderDecode`/`ZydisDecoderFullDecode`. Renaming `ZydisDecoderDecodeBuffer` -> `ZydisDecoderDecodeInstruction` actually makes sense!

src/Encoder.c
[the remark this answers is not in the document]
mappzor: By "dangerous" I wanted to highlight that new check does something different from the original check. I thought that change in behavior was unintentional. I agree there's no perfect solution here when it comes to handling operands, so `ZydisEncoderDecodedInstructionToEncoderRequest` will be no longer crash-proof as it used to be. I wanted to use `ZydisDecodedInstruction.operand_count_visible` for slightly cleaner API but on the other hand current signature is consistent with other functions that accept "a tuple" of operands and operand count. I don't have a strong opinion about this, so I guess it's ok as it is.

> We can add an additional check for e.g. ZYDIS_OPERAND_COUNT_MAX/ZYDIS_ENCODER_MAX_OPERANDS on these places, but even then it would access invalid memory in my example above.

I would totally bring back check for `ZYDIS_ENCODER_MAX_OPERANDS`. Ok, we cannot prevent user from crashing software with malformed operands but thanks to this check we can at least prevent crashes due to out of bounds accesses to encoder request (that's the intent behind original check).

mappzor:
> I like the idea of providing one function that combines instruction- and operand-decoding. However I would probably just name the utility function to something like `ZydisDecoderDecode`/`ZydisDecoderFullDecode`.

Both sound good to me. However if latter one is preferred I'd go with `ZydisDecoderDecodeFull` to stay consistent with `ZydisDecoderDecode(Instruction|Operands)`. It's important to document that change in porting guide and it's probably best place to mention that alternative approach to decoding exists in v4.

src/Encoder.c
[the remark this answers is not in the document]
flobernd: Perfect :-) Will reintroduce the upper bounds check on all places and as well change the `>` to `!=` (this part indeed was unintentional).

include/Zydis/Formatter.h
athre0z: Ah, sorry -- meant to comment that on line 1027 / 1046, but selected the wrong one when switching between different diffing views. This change wouldn't make sense for this one.

include/Zydis/DecoderTypes.h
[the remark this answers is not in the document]
athre0z: > I think we should discuss this in detail. 

Agreed, but this definitely deserves a dedicated issue for proper discussion -- let's not do that in this PR.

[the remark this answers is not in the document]
flobernd: > let's not do that in this PR

Definitely not! Will create a separate issue tomorrow.

include/Zydis/Decoder.h
```diff
@@ -223,8 +225,24 @@ ZYDIS_EXPORT ZyanStatus ZydisDecoderEnableMode(ZydisDecoder* decoder, ZydisDecod
...
+    ZydisDecoderContext* context, const void* buffer, ZyanUSize length, 
+    ZydisDecodedInstruction* instruction);
+
+/**
+ * Decodes the instruction operands.
+ *
+ * @param   decoder         A pointer to the `ZydisDecoder` instance.
+ * @param   context         A pointer to the `ZydisDecoderContext` struct.
+ * @param   instruction     A pointer to the `ZydisDecodedInstruction` struct.
+ * @param   operands        A pointer to the memory that receives the decoded operands.
```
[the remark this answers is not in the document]
flobernd: Let's do the documentation improvement in a separate step as well.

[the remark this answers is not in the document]
athre0z: Sounds good!

src/Decoder.c
```diff
@@ -1950,102 +1864,110 @@ static ZyanStatus ZydisDecodeOperands(ZydisDecoderContext* context,
...
+                                            (operands[i].mem.base == ZYDIS_REGISTER_BP))
+                                        {
+                                            operands[i].mem.segment = ZYDIS_REGISTER_SS;
+                                        }
+                                        else
+                                        {
+                                            operands[i].mem.segment = ZYDIS_REGISTER_DS;
+                                        }
+                                    }
+                                }
```
athre0z, line 1945: Not a fan of the rightward drift here, perhaps just going with inverted conditions and `goto`s here might be preferable? Could also do `do { .. } while (ZYAN_FALSE);` and break out of it, but that's essentially just a less honest goto.

src/Formatter.c
```diff
@@ -448,18 +448,20 @@ ZyanStatus ZydisFormatterSetHook(ZydisFormatter* formatter, ZydisFormatterFuncti
...
-    const ZydisDecodedInstruction* instruction, char* buffer, ZyanUSize length,
-    ZyanU64 runtime_address)
+    const ZydisDecodedInstruction* instruction, const ZydisDecodedOperand* operands,
+    ZyanU8 operand_count, char* buffer, ZyanUSize length, ZyanU64 runtime_address)
 {
-     return ZydisFormatterFormatInstructionEx(formatter, instruction, buffer, length,
-         runtime_address, ZYAN_NULL);
+     return ZydisFormatterFormatInstructionEx(formatter, instruction, operands, operand_count,
+         buffer, length, runtime_address, ZYAN_NULL);
 }
```
athre0z, line 456: Like already discussed in Discord: let's not have the `Ex` variant -- this one argument just isn't worth the complexity.

src/Encoder.c
```diff
@@ -2692,8 +2692,11 @@ ZyanBool ZydisCheckConstraints(const ZydisEncoderInstructionMatch *match)
 ZyanBool ZydisIsDefinitionCompatible(ZydisEncoderInstructionMatch *match,
     const ZydisEncoderRequest *request)
 {
-    const ZyanU8 def_op_count =
-        ZydisGetOperandDefinitions(match->base_definition, &match->operands);
+    match->operands = ZydisGetOperandDefinitions(match->base_definition);
+    const ZyanU8 def_op_count = match->base_definition->operand_count;
+
+    // TODO: Use new `operand_count_visible` field
```
athre0z, line 2698: Is this supposed to be resolved before merging or later?

src/Formatter.c
flobernd: Ah well the discussion was about the decoder "Ex" variant, which is already removed again.

The formatter variant has been around since the beginning. We can think about removing it in a separate change I guess.

src/Encoder.c
flobernd: I see some obvious optimizations, but I'm not sure if these are the ones @mappzor mentioned.

mappzor: Optimization opportunities are more about operand masks rather than `operand_count_visible`. However this new field allows for nice assert that would be hit if operand masks are wrong for whatever reason: `ZYAN_ASSERT(request->operand_count == match->base_definition->operand_count_visible)`. `if (def_op_count < request->operand_count)` check can be eliminated, visibility check if first `for` loop can be converted to assert and second loop can be removed completely.

flobernd: Did the suggested channges. Operand masks will be addressed by you in a later step?

mappzor: Hm, seems like fuzzers were able to hit that assert and operand masks should prevent that. Feel free to revert latest commit and just leave TODO as is. I will take care of it later.

flobernd: Done.

src/Formatter.c
athre0z: Ah yeah sorry, my bad, didn't properly read it and just wanted to tick mental point "get rid of Ex variant". Nevermind! That being said, the getting rid of the formatter `Ex` does seem desirable (even if not intentional by this comment :see_no_evil:).

src/Encoder.c
athre0z: Alright, thanks for investigating!

athre0z (approved):
LGTM! :-) One remaining open comment, but it's just style / not a blocker.

**Edit:** Ah, wait! We should also update the porting guide.

flobernd:
Will update the porting guide before merge.

Open points for later:

- Discussion: Feature `IFDEFS` and bitstructs in public API
> I think we should discuss this in detail. One good solution might be to introduce a ZYDIS_SIZE_OPTIMIZED_UNSAFE_FOR_BINDINGS option. This would indicate that Zydis is used in a pure C/C++ project and always built on demand (-> we know the feature defines of headers and code files are the same). This would as well allow us to e.g. change the ZydisDecodedOperand struct to unions or to introduce bit-fields in the public API.
- Discussion: Remove `ZydisFormatterFormatEx`
- Optimize encoder (operand masks)

PR #283, opened 2021-12-08
Add lookup tables for `ZydisRegister` and `ZydisRegisterClass`

### LINKED ISSUE #282: Refactor `ZydisRegisterGet(Id|Class)` to improve performance
Generating lookup tables for all valid `ZydisRegister` values has a noticable impact on encoder's performance as these two functions are used quite a lot during operand validation/encoding.

flobernd:
Closes #282


src/Generated/RegisterClassLookup.inc
```diff
@@ -0,0 +1,23 @@
+static const ZydisRegisterClassLookupItem REG_CLASS_LOOKUP[] =
+{
+    /* INVALID */ { -1, -1, ZYDIS_REGISTER_NONE, ZYDIS_REGISTER_NONE },
```
mappzor, line 3:
```suggestion
    /* INVALID */ { ZYDIS_REGISTER_NONE, ZYDIS_REGISTER_NONE, 0, 0 },
```

src/Generated/RegisterLookup.inc
```diff
@@ -0,0 +1,269 @@
...
+    /* K4         */ { ZYDIS_REGCLASS_MASK, 4, 64, 64 },
+    /* K5         */ { ZYDIS_REGCLASS_MASK, 5, 64, 64 },
+    /* K6         */ { ZYDIS_REGCLASS_MASK, 6, 64, 64 },
+    /* K7         */ { ZYDIS_REGCLASS_MASK, 7, 64, 64 },
+    /* BND0       */ { ZYDIS_REGCLASS_BOUND, 0, 128, 128 },
+    /* BND1       */ { ZYDIS_REGCLASS_BOUND, 1, 128, 128 },
+    /* BND2       */ { ZYDIS_REGCLASS_BOUND, 2, 128, 128 },
+    /* BND3       */ { ZYDIS_REGCLASS_BOUND, 3, 128, 128 },
+    /* BNDCFG     */ { ZYDIS_REGCLASS_BOUND, -1, 0, 0 },
+    /* BNDSTATUS  */ { ZYDIS_REGCLASS_BOUND, -1, 0, 0 },
```
mappzor, line 264: This is a change from previous version. `BNDCFG` and `BNDSTATUS` used to report `ZYDIS_REGCLASS_INVALID`, so new version causes fuzzers to crash. It's also inconsistent with `ZydisRegisterClassLookupItem` which reports correct `BND0-BND3` range.

I will compare this implementation against previous one to find other possible changes.

tools/ZydisInfo.c
```diff
@@ -1319,6 +1319,11 @@ int main(int argc, char** argv)
     ZYAN_PUTS("");
     PrintSegments(&instruction, &data[0]);
 
+    ZydisEncoderRequest request;
+    ZyanUSize len = instruction.length;
+    status = ZydisEncoderDecodedInstructionToEncoderRequest(&instruction, operands, instruction.operand_count_visible, &request);
+    status = ZydisEncoderEncodeInstruction(&request, data, &len);
+
```
mappzor, line 1326: Leftover from debugging?

src/Generated/RegisterLookup.inc
mappzor: Here's a summary of all the breaking changes:
| What | Old | New | Opinion |
| --- | --- | --- | --- |
| RAX-R15 width16/32 | 0 | 64 | :x: |
| X87CONTROL, X87STATUS, X87TAG | ZYDIS_REGCLASS_INVALID | ZYDIS_REGCLASS_X87 | :x: |
| GDTR, LDTR, IDTR, TR | ZYDIS_REGCLASS_INVALID | ZYDIS_REGCLASS_TABLE | :heavy_check_mark: |
| K0-K7 width16/32/64 | 0 | 64 | :question: |
| BNDCFG, BNDSTATUS | ZYDIS_REGCLASS_INVALID | ZYDIS_REGCLASS_BOUND | :x: |

:x: - bringing old behavior back is desirable
:heavy_check_mark: - new behavior is desirable
:question: - change is good but mask registers are 16-bits on KNC and they share same exact names with AVX-512 64-bits registers. If we want to keep this change then TODO: Stop treating `ZYDIS_REGCLASS_MASK` as special case inside `ZydisIsRegisterOperandCompatible` (correct this check: `if (reg_width == 0 && (reg_class != ZYDIS_REGCLASS_MASK))`).

tools/ZydisInfo.c
flobernd: Whoops. Yeah. We need a simple command line tool like `ZydisInfo` which does re-encoding to debug specific byte sequences.

Might not make super much sense, but maybe we can even add "size optimized" (re-encide) as a new info output to `ZydisInfo`.

src/Generated/RegisterLookup.inc
flobernd: Thanks for checking. Makes sense! I have to adjust my simple python script to reflect the suggested changes.

flobernd: One condition was missing in the new Python script which I had in the old Delphi based one.

flobernd: > but mask registers are 16-bits on KNC

Right. Forgot about this. This was the reason for the `0`. I think `64` still makes more sense as KNC essentially is dead.

mappzor: > Right. Forgot about this. This was the reason for the `0`. I think `64` still makes more sense as KNC essentially is dead.

I agree, it's better to prioritize most common case rather than return 0.

flobernd: I just pushed some changes. Just removed the second part of the check inside `ZydisIsRegisterOperandCompatible` and left the `if (reg_width == 0)` like it was.

mappzor: LGTM :)

PR #284, opened 2021-12-09
Reworked encoding of branching instructions (Fixes #268, Fixes #266)

### LINKED ISSUE #266: Cant encode using ZYDIS_ENCODABLE_BRANCH_TYPE_NEAR32
```cpp
ZydisEncoderRequest req{};
req.mnemonic = ZYDIS_MNEMONIC_JMP;
req.branch_type = ZydisEncodableBranchType::ZYDIS_ENCODABLE_BRANCH_TYPE_NEAR32;
req.operand_count = 1;
req.operands[0].type = ZYDIS_OPERAND_TYPE_IMMEDIATE;
req.operands[0].imm.u = 0xFF;
```
When specifying NEAR32 I sort of expected it to encode ```jmp <rel32>``` considering 0xFF is in range but it appears I have to either specify ZYDIS_ENCODABLE_BRANCH_TYPE_NEAR64 or ZYDIS_ENCODABLE_BRANCH_TYPE_SHORT.

Also to my understanding there is no encoding variant that encodes 64 bit relative address.

### LINKED ISSUE #268: Consider reworking branch types in the encoder
This is a continuation of the discussion [here](https://github.com/zyantific/zydis/pull/254#discussion_r739544494). Thinking about this for a bit longer and having gained some usage experience with the encoder, I withdraw my initial suggestion. 

When encoding branches, a user must currently explicitly choose a very specific branch type like `ZYDIS_ENCODABLE_BRANCH_TYPE_NEAR64` in the request, which is quite inconvenient for code generation. I think that it would be preferable to instead keep the decoder type as-is, and instead split the `ZydisEncodableBranchType` type into two separate fields in the encoder request such as:

```c
enum ZydisBranchWidth {
  ZYDIS_BRANCH_WIDTH_NONE,
  ZYDIS_BRANCH_WIDTH_16,
  ZYDIS_BRANCH_WIDTH_32,
  ZYDIS_BRANCH_WIDTH_64
};

struct ZydisEncoderRequest {
  ZydisBranchType branch_type; // reusing the decoder type
  ZydisBranchWidth branch_width;
  // remaining fields omitted
};
```

`ZYDIS_BRANCH_TYPE_NONE` and `ZYDIS_BRANCH_WIDTH_NONE` would mean "infer" to the encoder. `ZYDIS_BRANCH_TYPE_NONE` specifically would mean "infer whether near or short", but never consider "far", which must be requested explicitly. I'd argue that in 99% of cases, users will want some variation of a near/short branch. In a semantic encoder interface, I'd further argue that it should be the encoder that picks a good branch variant for the provided offset.

The upside with the current approach is that it's impossible to specify invalid combinations, reducing the amount of validation required. I assume that's what you meant with "needing a tuple type" in your comment [here](https://github.com/zyantific/zydis/pull/254#discussion_r739657745). However, I personally don't think this warrants the downsides. If such a tuple type is truly required, we should keep it internal to the encoder, abstracting it away from the user.

mappzor:
Fixes #268
Fixes #266

### DIFF
--- a/examples/EncodeFromScratch.c
+++ b/examples/EncodeFromScratch.c
@@ -85,7 +85,6 @@ static ZyanUSize AssembleCode(ZyanU8* buffer, ZyanUSize buffer_length)
     // Assemble `ret`.
     memset(&req, 0, sizeof(req));
     req.mnemonic = ZYDIS_MNEMONIC_RET;
-    req.branch_type = ZYDIS_ENCODABLE_BRANCH_TYPE_NEAR64;
     req.machine_mode = ZYDIS_MACHINE_MODE_LONG_64;
     AppendInstruction(&req, &write_ptr, &remaining_length);
 
--- a/include/Zydis/Encoder.h
+++ b/include/Zydis/Encoder.h
@@ -112,29 +112,26 @@ typedef enum ZydisEncodableEncoding_
 } ZydisEncodableEncoding;
 
 /**
- * Defines encodable branch types.
+ * Defines encodable physical/effective sizes of relative immediate operands. See
+ * `ZydisEncoderRequest.branch_width` for more details.
  */
-typedef enum ZydisEncodableBranchType_
+typedef enum ZydisBranchWidth_
 {
-    ZYDIS_ENCODABLE_BRANCH_TYPE_NONE,
-    ZYDIS_ENCODABLE_BRANCH_TYPE_SHORT,
-    ZYDIS_ENCODABLE_BRANCH_TYPE_NEAR16,
-    ZYDIS_ENCODABLE_BRANCH_TYPE_NEAR32,
-    ZYDIS_ENCODABLE_BRANCH_TYPE_NEAR64,
-    ZYDIS_ENCODABLE_BRANCH_TYPE_FAR16,
-    ZYDIS_ENCODABLE_BRANCH_TYPE_FAR32,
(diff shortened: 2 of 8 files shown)


flobernd:
LGTM! 

New API looks nice and documentation is well done. 

Just roughly swiped over the code details and noticed the special case handling for certain instructions by checking their mnemonics. While I don't really like that approach, I don't have a better solution in mind for that right now (besides adding new flags ofc. 😋). It might cause complications if - at some point - we can finally allow the user to choose what kind of extensions should be build into the binary (using CMake options). But until then, it's okay for me to  leave the hardcoded checks like they are right now.

mappzor:
Yeah, I didn't like hardcoding `XBEGIN` AGAIN but we don't really have much choice. This is a very unique instruction at the moment and we already have a special case for it in the formatter.

With bugged `MVEX`instructions (actually they are more like one instruction with slightly different variants, they even have same opcode just different `REX.W` and `MVEX.EH`) I decided to hardcode because it's dead hardware with only a single offender. Every other solution seemed to be a massive overkill.

flobernd (approved):

flobernd:
LGTM from my side, maybe @athre0z want's to have a look as well.

athre0z (approved):
Awesome changes, thanks a lot for that! Definitely makes for a much better UX. I'll merge this later after I pushed the corresponding new fuzzing corpus.

PR #290, opened 2021-12-16
Improved prefix handling

mappzor:
This PR fixes a bunch of prefix-related issues:

- [Decoder] Branch hints were not decoded in 64-bit mode
- [Encoder] `XACQUIRE`/`XRELEASE` support was broken
- [Encoder] Multiple REP-like prefixes could be specified simultaneously
- [Encoder] `NOTRACK` and segment overrides could be specified simultaneously when mode=!64 (now this is enforced in all modes to match decoder's behavior)
- [Encoder] Fuzzer now verifies if encoded prefixes match encoder request


flobernd (approved):
LGTM! Thanks.

PR #298, opened 2021-12-22
Fix `MASM` style disassembly

### LINKED ISSUE #136: Wrong hex formatting in MASM formatter
In MASM hex constants are not allowed to begin with a character.

Sample disassembly with MASM formatter:
```asm
push qword ptr [rbp+Ch]
```

Correct masm compatible output should look like this:
```asm
push qword ptr [rbp+0Ch]
```

flobernd:
Closes #136


athre0z (dismissed):
LGTM, thanks for tackling that! :) Sorry for the late review -- needed a few days off from code.

PR #300, opened 2022-01-06
Improve CMake file

flobernd:
## Change dependency path from "per project" to "per dependency"

`ZYDIS_ZYCORE_PATH` -> `ZYAN_ZYCORE_PATH`
...

This way we do not need to change multiple path variables when a dependency is used multiple times. E.g. in case of `Zyrex`:

- Zyrex
  - Zycore (`./dependencies/zycore` [`ZYREX_ZYCORE_PATH`])
  - Zydis (`./dependencies/zydis` [`ZYREX_ZYDIS_PATH`])
    - Zycore (`./dependencies/zydis/dependencies/zycore` [`ZYDIS_ZYCORE_PATH`])

After my change there is only one `Zycore` path pointing to the top-level dependency:

- Zyrex
  - Zycore (`./dependencies/zycore` [`ZYAN_ZYCORE_PATH`])
  - Zydis (`./dependencies/zydis` [`ZYAN_ZYDIS_PATH`])
    - Zycore (`./dependencies/zycore` [`ZYAN_ZYCORE_PATH`])

Only negative point: This prevents the user to intentionally choose different dependency versions for individual sub-dependencies. I don' think this is really needed tho and as well to fully support that (at least for Windows shared builds) we would have to add a version suffix to the generated `.dll` files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,10 +49,12 @@ option(ZYDIS_FUZZ_AFL_FAST
 option(ZYDIS_LIBFUZZER
     "Enables LLVM libfuzzer mode and reduces prints in ZydisFuzzIn"
     OFF)
-option(ZYDIS_SYSTEM_ZYCORE
+
+# Dependencies
+option(ZYAN_SYSTEM_ZYCORE
     "Use system Zycore library"
     OFF)
-set(ZYDIS_ZYCORE_PATH
+set(ZYAN_ZYCORE_PATH
     "${CMAKE_CURRENT_LIST_DIR}/dependencies/zycore"
     CACHE
     PATH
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -6,6 +6,16 @@ include(GenerateExportHeader)
...
+if (NOT DEFINED CMAKE_ARCHIVE_OUTPUT_DIRECTORY)
+    set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+endif()
+if (NOT DEFINED CMAKE_LIBRARY_OUTPUT_DIRECTORY)
+    set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+endif()
+if (NOT DEFINED CMAKE_RUNTIME_OUTPUT_DIRECTORY)
+    set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+endif()
+
```
athre0z, line 18: Hijacking global CMake variables is strongly discouraged for libraries in general. It can lead to situations where you just add a library and then your whole build starts behaving differently, since these are global. Instead, use `set_target_property` with `LIBRARY_OUTPUT_DIRECTORY` etc. to do this on a per-target basis.

athre0z (commented):
The `ZYDIS` -> `ZYAN` rename for re-usability looks good to me. Will break the Debian build though (CC @Tachi107).

CMakeLists.txt
Tachi107: @athre0z don't worry, before updating I always check if I have to change the Debian build script, but thanks for keeping me updated :)

Tachi107:
> The `ZYDIS` -> `ZYAN` rename for re-usability looks good to me. Will break the Debian build though (CC @Tachi107).

@athre0z don't worry, before updating I always check if I have to change the Debian build script, but thanks for keeping me updated :)

CMakeLists.txt
flobernd: I removed that part for now. Will create a separate PR after doing some research.

PR #302, opened 2022-01-07
Add script for generating an amalgamated distribution

athre0z:
This PR adds a script that creates an amalgamated build of Zydis. In other words: the script merges all sources files and internal headers into a single `.c` file and all public headers into a single `.h` files. The term "amalgamated" was inspired by libraries like SQLite that provide similar distributions.

~~I'll add a CI build (and perhaps automate attaching such a build to releases) in a separate PR.~~ (amended to this PR)

~~Note: this PR is based on the `remove-generateexportheader` branch, because the export config was previously causing problems in the amalgamated builds.~~ (merged and rebased on master)


src/Generated/EnumRegister.inc
```diff
@@ -1,4 +1,4 @@
-static const ZydisShortString STR_REGISTER[] =
+static const ZydisShortString STR_REGISTERS[] =
```
athre0z, line 1: This name clashes with another `STR_REGISTER` somewhere in the generated formatter strings when doing amalgamated builds, so I had to rename it. I'll push a commit to also change that in `gen_registers.py` once this PR is merged.

flobernd (approved):
LGTM!

athre0z:
Figured that it probably makes more sense to also add CI in this PR, so added a commit for that. Turned out that the CI's Python version isn't latest, so I had to downgrade a few things. Will clean up history after approval.

flobernd (approved):

Tachi107:
> Turned out that the CI's Python version isn't latest, so I had to downgrade a few things.

In case you didn't know, GitHub's CI offers the possibility to run jobs in Docker images, so you're not limited to using Ubuntu LTS; if you need updated packages you could use Debian Testing, it's really nice :)

Here's a quick example (taken from [Tachi107/cloudflare-ddns/.github/workflows/muon.yam](https://github.com/Tachi107/cloudflare-ddns/blob/c69210e6411d0b725c20a930eb066a7b49cf8260/.github/workflows/muon.yaml)):

```yaml
jobs:
  muon:
    runs-on: ubuntu-latest
    container:
      image: debian:testing

    steps:
      ...build steps...
```

athre0z:
That's good to know, thanks!

PR #307, opened 2022-01-31
Fixed newlines on Windows (amalgamation)

mappzor:
When opening files in `r`/`w` modes Python performs `CRLF <-> LF` conversions automatically. Using `os.linesep` causes `LF` inside `CRLF` to be converted, resulting in `CRCRLF`.


athre0z (approved):
Ah, good to know! Thanks! :-)

PR #308, opened 2022-02-06
Fixed edge case for hidden operand scaling (Fixes #296)

### LINKED ISSUE #296: Encoder may prepend operand size prefix to reduce length
It seems possible that for some instructions the encoder might add an operand size prefix to shorten the immediate this however changes usually how the instruction works, a concrete example would be ```push <imm>```.

Encoding ```push 0x1CC``` will generate following bytes:
```
66:68 CC01 | push 1CC
```` 
I think the encoder should not prepend prefixes in order to shrink the size, this potentially changes how the instruction operates. Other instructions other than push might be affected, can't think of any right now.

mappzor:
Fixes #296 

This problem turned out to affect `push imm` only because it fulfilled 3 conditions at the same time:

- scales with hidden operand size
- has `imm` operand
- belongs to `default64` operand size map

Last condition is the major reason why in 64-bit mode 16-bit size was preffered (other modes were fine). The logic inside `ZydisGetOperandSizeFromElementSize` is correct though and required by other instructions. This fix corrects problem by changing logic inside `ZydisGetOszFromHint`. Address size scaling doesn't seem to require any modifications.

### DIFF
--- a/src/Encoder.c
+++ b/src/Encoder.c
@@ -311,45 +311,29 @@ ZyanU8 ZydisGetMachineModeWidth(ZydisMachineMode machine_mode)
  */
 ZyanU8 ZydisGetAszFromHint(ZydisAddressSizeHint hint)
 {
-    switch (hint)
-    {
-    case ZYDIS_ADDRESS_SIZE_HINT_NONE:
-        return 0;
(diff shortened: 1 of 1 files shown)


flobernd (approved):
LGTM! Thanks :-)

PR #324, opened 2022-03-22
Cleanup zydis include

serge-sans-paille:
Remove unused headers / replace them by a selection if needs be.
Cleanup assisted by iwyu[0]
This results in slightly less lines in the preprocessed output (-6908) but also
reduces coupling between compilation units.

[0] https://include-what-you-use.org/


serge-sans-paille:
Thanks for running the test, all green now \o/ Anything else I can do for this PR ?

athre0z (approved):
Looks good to me, thanks! :-)

PR #326, opened 2022-04-20
Add unions in decoded instruction and operands

athre0z:
This PR adds anonymous unions to the operand struct and the vector-extension structs within the `raw` struct, reducing the struct size quite substantially. For the operands, I also moved the `type` field immediately in front of the union, which should allow us to represent it as a safe union in the Rust bindings without need for calling an accessor.

Anonymous unions are a C11 feature, so I had to bump the "required compiler standard" in the process. Because MSVC doesn't properly support C11, we instead rely on the non-standard extension here, silencing the corresponding warning.


flobernd (approved):
LGTM!

PR #328, opened 2022-04-23
Enforce consistent file encodings

athre0z:
This PR adjusts the `.gitattribute` to enforce both consistent line endings and file encodings. Thanks to @bjorn3 for suggesting this approach!


flobernd (approved):
LGTM!

PR #332, opened 2022-05-01
Encoder: make all internal functions `static`

athre0z:
As pointed out by @ZehMatt, the encoder previously didn't have the internal functions marked as `static`. Particularly on *nix platforms that support semantic interposition, this makes a huge different in terms of performance. In a quick and dirty tests, it appears to be accelerating the encoder by about 20% when built with GCC.

CC @mappzor


flobernd (approved):
LGTM!

PR #334, opened 2022-05-01
Encoder: improve `ZydisArePrefixesCompatible` perf

athre0z:
Most instructions tend to not have any custom prefixes, so it greatly improves overall encoder performance when we check if any prefix is requested at all and potentially early-exit before doing a detailed check.

In a small benchmark, this improved overall encoder performance by about 10%.

CC @mappzor

### DIFF
--- a/src/Encoder.c
+++ b/src/Encoder.c
@@ -2762,6 +2762,12 @@ static ZyanBool ZydisIsDefinitionCompatible(ZydisEncoderInstructionMatch *match,
  */
 static ZyanBool ZydisArePrefixesCompatible(const ZydisEncoderInstructionMatch *match)
 {
+    // Early-exit optimization for when no prefixes are requested at all.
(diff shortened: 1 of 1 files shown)


src/Encoder.c
```diff
@@ -2762,6 +2762,25 @@ static ZyanBool ZydisIsDefinitionCompatible(ZydisEncoderInstructionMatch *match,
  */
 static ZyanBool ZydisArePrefixesCompatible(const ZydisEncoderInstructionMatch *match)
 {
+    static const ZyanU64 all_checked_prefixes =
```
mappzor, line 2765: `ZYDIS_ENCODABLE_PREFIXES` can be used instead.

mappzor (changes requested):

flobernd (approved):
LGTM!

PR #338, opened 2022-05-07
Extend porting guide

athre0z:
In light of the v4.0 release drawing closer, I took some time to refine our porting guide, putting in everything that I remember and also doing a quick sweep through the git-blame, looking for recent changes.


flobernd (approved):
LGTM!

PR #340, opened 2022-05-08
Duplicate encoding to aid Rust bindings

athre0z:
As previously discussed in Discord, this PR duplicates the encoding field in the `raw` section of the decoded instruction to allow the Rust bindings to model it as a safe enum.

### DIFF
--- a/include/Zydis/DecoderTypes.h
+++ b/include/Zydis/DecoderTypes.h
@@ -1143,6 +1143,13 @@ typedef struct ZydisDecodedInstruction_
         } rex;
 
         /*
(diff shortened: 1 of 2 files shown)


flobernd (approved):
LGTM!

I just noticed the `rex` field is missing from the union. `REX` is illegal if either a `XOP/VEX/EVEX` or `MVEX` prefix is present (or visa verce). The decoder should return `ZYDIS_STATUS_ILLEGAL_REX` in this case.

flobernd (approved):
LGTM!

PR #341, opened 2022-05-09
Always initialize `instruction->cpu_flags/fpu_flags`

### LINKED ISSUE #319: Always assign ZydisDecodedInstruction::cpu_flags
For instructions that don't use any flags this field is set to to nullptr. Ideally this just points to an empty struct for those so the user doesn't need to check if this is valid memory.

flobernd:
Closes #319


athre0z (approved):
Ah, so it didn't require generator changes after all? Nice.

flobernd:
> Ah, so it didn't require generator changes after all? Nice.

Yes, the generator already emits a fallback value as the first element in the list.

PR #348, opened 2022-06-06
Move segment functions into a separate, optional header

athre0z:
This PR splits the segment functions from `Utils.h` into a separate header and source file and makes it an optional feature. 

Most users won't need this API, and with the intention to extend this API to essentially replace most of what was previously provided via the `raw` instruction part, I think it makes sense to decouple it.


flobernd (approved):
LGTM!

Issue #354, opened 2022-06-23
Windows Wsign-compare warning in SharedData.c


rvandermeulen:
This is with v3.2.1, but as far as I can tell, the code is the same on tip. We can suppress the warning if this isn't something you're inclined to fix, but filing in case you are 😄 
```
SharedData.c(165,25): error: comparison of integers of different signs: 'ZydisInternalElementType' (aka 'enum ZydisInternalElementType_') and 'unsigned int' [-Werror,-Wsign-compare]
    ZYAN_ASSERT(element < ZYAN_ARRAY_LENGTH(lookup));
    ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~
Defines.h(268,42): note: expanded from macro 'ZYAN_ASSERT'
#   define ZYAN_ASSERT(condition) assert(condition)
                                  ~~~~~~~^~~~~~~~~~
windows kits/10/include/10.0.17134.0/ucrt/assert.h(33,17): note: expanded from macro 'assert'
            (!!(expression)) ||                                                              \
                ^~~~~~~~~~
1 error generated.
```

athre0z:
We can definitely fix that on our side, no need to suppress the warning. Out of curiosity, what is the build setup that you're using there? You mention that you're on Windows, but the compiler arguments are printed with dashes (`-Werror` etc.), so probably MinGW? We should probably have something like this in our CI as well.

rvandermeulen:
This was with clang-cl 14.0.5

PR #357, opened 2022-07-02
Add ClangCL build to CI + fix sign compare warning

athre0z:
This PR does a bunch of minor clean-up to our CI config, adds a dev-mode clang-cl build and finally fixes the sign compare warning that was reported in #354. I recommend reviewing commit-by-commit.

In case you're wondering why this branch is on a fork: I originally wanted to dev this on a fork to prevent mail-flooding while I brute-force the CI config, but accidentally created the PR here instead of on the fork.


.github/workflows/main.yml
```diff
@@ -1,4 +1,4 @@
-name: GitHub Actions CI
+name: CI
```
flobernd, line 1: You need to change that name in the README batch as well, otherwise it won't be displayed correctly.

flobernd: And we should probably do the same changes in the `zycore-c` repo.

athre0z: > You need to change that name in the README file as well (GitHub actions batch), otherwise it won't be displayed correctly.

Did that now. It's 404ing for the moment -- either this refers to a different name, or it only actually applies when merged on master. Guess we'll see on merge.

flobernd: Yeah it's always referring to master. Will be good after the merge :-)

flobernd (approved):

athre0z:
I just noticed that our previous configuration was built on incorrect assumptions about [how `include` works in GitHub actions](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations). Various builds that we wanted to run were not actually executed (e.g. minimal-mode builds), and multiple additional ones that we didn't want were.

For our purposes it's a lot simpler to reason about when just writing out the full list of tests to run manually. Also simplifies the actual build `steps` by requiring a lot fewer variables.

flobernd:
@athre0z 

As an alternative, you could check the actions in the zycore-c repository. I "recently" updated these ones to run a nice "cross-product" of compiler/arch/nolibc builds.

Btw. what was the reason to include dev-mode in the CI? Didn't that just change the way CMake adds the source/header files to the generated (VS-) project file?

athre0z:
> As an alternative, you could check the actions in the zycore-c repository. I "recently" updated these ones to run a nice "cross-product" of compiler/arch/nolibc builds.

That works well for Zycore, but it's not so simple for Zydis. Essentially, this part of our config could easily be expressed in a similar manner:

https://github.com/athre0z/zydis/blob/7f71301c79b6142615bdaeb411e471c2ee444c35/.github/workflows/main.yml#L18-L30

But we additionally want a few more specialized / non-standard builds to be run:

https://github.com/athre0z/zydis/blob/7f71301c79b6142615bdaeb411e471c2ee444c35/.github/workflows/main.yml#L33-L39

These, we don't necessarily want to run on every platform: it'd result in an explosion of combinations. We'd go from 6 to 24 configurations. For the `-TClangCL` rule, I'm not even sure if that is expressible at all: we only want that to be added for Windows, but also not always.

I originally assumed that this is what `include` is meant for: define most stuff via a Cartesian product, then add a few additional configurations. However, that's not at all what it does. It instead matches those variables that also exist in the matrix and then adds the variables that don't exist in the matrix for each of those. 

> Btw. what was the reason to include dev-mode in the CI? Didn't that just change the way CMake adds the source/header files to the generated (VS-) project file?

It also enables `-Werror -Wall -Wpedantic` and allows us to catch warnings in CI.

flobernd:
@athre0z

Alright, makes sense for the first part.

> It also enables -Werror -Wall -Wpedantic and allows us to catch warnings in CI.

Ah, right. In this case I would even suggest to always enable DEV_MODE in the CI for all of the build configs.

athre0z:
> Ah, right. In this case I would even suggest to always enable DEV_MODE in the CI for all of the build configs.

Good point -- can't think of a reason not do that, will change!

flobernd (approved):

PR #361, opened 2022-07-10
Simplify formatter interface

athre0z:
This PR removes the `Ex` variants of various formatter functions. Their functionality is instead integrated into the non-`Ex` variant. Having the `Ex` variants effectively doubled the complexity of our public formatter interface and each only had a single additional argument. This additional argument can trivially be passed with as `NULL`, which I believe should be intuitive to most seasoned C programmers.

When porting the various examples I varied between using `NULL` and `ZYAN_NULL`. I chose one or the other depending on whether the example otherwise prefers Zycore or standard C constructs. We should probably normalize that across all examples at some point.


flobernd (approved):
LGTM!

PR #364, opened 2022-07-11
Updating ZyCore

matthew-olson-intel:
This pulls in the latest ZyCore updates, such as the recent `CMakeLists.txt` addition, so that Zydis can compile on Amazon Linux.


athre0z:
It looks like you attempted to update it to a commit that doesn't actually exist. You'll have to checkout `master` in the submodule: you probably accidentally attempted to pin the submodule to a commit that only exists on your fork.

PR #371, opened 2022-07-29
Formatter: make `operand_count` check less strict

athre0z:
This PR changes the `operand_count` check to be less strict, allowing the user to simply pass `sizeof(operand_array)` as the operand count. The count is passed exclusively as an additional sanity check for the user and I find the new behavior much more intuitive. If I see a size argument following a pointer argument, I automatically assume it to receive the array length.


flobernd:
Should be fine this way! It does not match the behavior of `ZydisDecoderDecode*` now, but at the end they are doing two completely different things anyways.

flobernd (approved):

PR #372, opened 2022-07-29
Add README examples to our build

athre0z:
We have a tendency of forgetting to update the README examples due to them not being included in the build. This PR fixes that and instead embeds the examples [via the code-browsers that GitHub renders for permalinks](https://github.com/zyantific/zydis/blob/597e8bbcf408bae8bbaf3ced57cc992955dba6e5/README.md#decoder-example). We'll unfortunately still have to update the perma-links, but having changes for the corresponding files in a PR will hopefully remind us of updating these. I accidentally pushed one commit from this PR to master directly (9cb54996c215422a398d7d2a287a08a185344200). If this PR is not accepted, I'll do a revert commit.


flobernd:
I like that very much! Would be really cool if we could use a relative path somehow. Did you do research in that direction?

athre0z:
I'm afraid that's not going to be possible: this [only works with permalinks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet#linking-to-code). It makes sense from an "under the hood" standpoint: this is the only way that they can easily cache this without making cache invalidation a huge effort.

flobernd:
Yeah I guessed so as well. Anyways, it's still a good improvement :-)

PR #375, opened 2022-08-10
Doc: group important macros and refer to them

athre0z:
I noticed that flag fields that use macros to define their values are currently very tough to navigate in our documentation. This PR resolves this by manually grouping them and referring to the respective groups. There's probably more macro groups that could profit from this -- I only covered the most important ones here.

### DIFF
--- a/include/Zydis/DecoderTypes.h
+++ b/include/Zydis/DecoderTypes.h
@@ -268,6 +268,15 @@ typedef struct ZydisDecodedOperand_
 typedef ZyanU32 ZydisAccessedFlagsMask;
 
 /**
+ * @defgroup decoder_cpu_flags CPU flags
+ * @ingroup decoder
+ *
+ * Constants used for testing CPU flags accessed by an instruction.
+ *
+ * @{
+ */
+
+/**
  * Carry flag.
  */
 #define ZYDIS_CPUFLAG_CF    (1ul <<  0)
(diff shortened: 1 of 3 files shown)


florian-sneller (approved):
Good improvement!

I would probably leave out the „macro“ suffix, but I guess that’s required to avoid overlapping names?

flobernd:
Good improvement!

I would probably leave out the „macro“ suffix, but I guess that’s required to avoid overlapping names?

flobernd (approved):

athre0z:
> I would probably leave out the „macro“ suffix, but I guess that’s required to avoid overlapping names?

Not really -- the group name doesn't have the "Zydis" prefix, so it doesn't clash regardless. Just thought it'd make it clearer what the group actually, well, groups. I can throw it out though, no strong opinion here.

flobernd:
> I can throw it out though, no strong opinion here.

No strong opinion here as well 😄

athre0z:
Ok: I made the CPU and FPU flags sub-groups of the decoder, the instruction attributes into a top-level group (because they are used in both encoder and decoder) and for the version stuff I simply extended the existing group to include the macros as well.

Screen Shot 2022-08-15 at 12 19 41

flobernd (approved):
LGTM!

PR #376, opened 2022-08-13
Added `ZydisEncoderNopFill`

mappzor:
Added small helper function to fill buffer with nops. Useful for padding, "erasing" code, etc.

### DIFF
--- a/src/Encoder.c
+++ b/src/Encoder.c
@@ -4706,4 +4706,37 @@ ZYDIS_EXPORT ZyanStatus ZydisEncoderDecodedInstructionToEncoderRequest(
     return ZYAN_STATUS_SUCCESS;
 }
 
+ZYDIS_EXPORT ZyanStatus ZydisEncoderNopFill(void *buffer, ZyanUSize length)
+{
+    if (!buffer)
+    {
+        return ZYAN_STATUS_INVALID_ARGUMENT;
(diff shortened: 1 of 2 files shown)


athre0z (approved):
LGTM, thanks!

src/Encoder.c
```diff
@@ -4706,4 +4706,30 @@ ZYDIS_EXPORT ZyanStatus ZydisEncoderDecodedInstructionToEncoderRequest(
     return ZYAN_STATUS_SUCCESS;
 }
 
+ZYDIS_EXPORT void ZydisEncoderNopFill(void *buffer, ZyanUSize length)
+{
```
flobernd, line 4710: We should add the "invalid argument" checks here for consistency (`buffer == null`, `length < 0`).

athre0z: Length == 0 is a legitimate input for this function IMO. It should just do nothing in this case.

flobernd: Yes, but `length > 0` should error imho.

athre0z: This case is already eliminated by the argument being unsigned, unless I’m missing something.

flobernd: Ah yes, true. So just the buffer I guess :p

PR #378, opened 2022-09-03
Decoder: unnest all structs in `ZydisDecodedInstruction`

athre0z:
Nested structs make the documentation very hard to navigate because they are displayed in `A::B::C` fashion which -- given our long struct names -- quickly escalates into unreadability:

Screenshot_2022-09-03_16-42-37

This PR unnests all remaining nested structures in `ZydisDecodedInstruction`. Given that we recently already started unnesting parts of this struct to allow for using anonymous unions, I think doing this for all sub-structs is only consistent.


flobernd (approved):
LGTM!

PR #380, opened 2022-09-12
Decoder: simplify ZydisDecoderDecodeFull API

athre0z:
As previously discussed in https://github.com/zyantific/zydis/pull/379, this PR simplifies `ZydisDecoderDecodeFull` in order to reduce the overall complexity of our public interface. If people wish to decode less than `ZYDIS_MAX_OPERAND_COUNT` operands, they can simply use the more advanced `ZydisDecoderDecodeInstruction` and `ZydisDecoderDecodeOperands` functions.

### DIFF
--- a/src/Decoder.c
+++ b/src/Decoder.c
@@ -4967,19 +4967,16 @@ ZyanStatus ZydisDecoderEnableMode(ZydisDecoder* decoder, ZydisDecoderMode mode, 
 
 ZyanStatus ZydisDecoderDecodeFull(const ZydisDecoder* decoder,
     const void* buffer, ZyanUSize length, ZydisDecodedInstruction* instruction,
-    ZydisDecodedOperand* operands, ZyanU8 operand_count, ZydisDecodingFlags flags)
+    ZydisDecodedOperand operands[ZYDIS_MAX_OPERAND_COUNT])
 {
-    if (!decoder || !instruction || (operand_count && !operands) ||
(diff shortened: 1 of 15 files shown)


src/Decoder.c
```diff
@@ -4987,19 +4986,14 @@ ZyanStatus ZydisDecoderDecodeFull(const ZydisDecoder* decoder,
...
+    ZYAN_CHECK(ZydisDecoderDecodeOperands(decoder, &context, instruction, operands,
+        instruction->operand_count));
 
-    const ZyanBool visible_only = flags & ZYDIS_DFLAG_VISIBLE_OPERANDS_ONLY;
-    const ZyanU8 count = visible_only
-        ? instruction->operand_count_visible
-        : instruction->operand_count;
+    ZYAN_ASSERT(instruction->operand_count <= instruction->operand_count);
+    ZYAN_MEMSET(&operands[instruction->operand_count], 0,
+        (ZYDIS_MAX_OPERAND_COUNT - instruction->operand_count) * sizeof(operands[0]));
```
athre0z, line 4994: I figured that it'd be rather unintuitive for this function not to zero the remaining entries now that it expects an array of a fixed size, but I'm open to changing this.

flobernd: I think we should keep it like this. It's consistent with the "simple disassembler API" where the complete struct is nulled. If performance is a concern, we can refer to the advanced functions as well.

flobernd (approved):

PR #386, opened 2022-10-09
Decoder: consistent handling of NULL buffer

athre0z:
I realized that all other decoder function respond to `buffer` being passed as `NULL` with returning `ZYDIS_STATUS_NO_MORE_DATA` and figured that the disassembler API should probably do the same for consistency.


flobernd (approved):
LGTM!

Just thinking we should probably change that behavior in the future and always return `INVALID_ARGUMENT` for a `null` buffer. Makes more sense to me as a `null` buffer should never get passed in this case.

athre0z:
> Just thinking we should probably change that behavior in the future and always return INVALID_ARGUMENT for a null buffer. Makes more sense to me as a null buffer should never get passed in this case.

Agreed. Shall I just re-purpose this PR to do just that?

flobernd:
> Shall I just re-purpose this PR to do just that?

Sure, let's do that 🙂

flobernd (approved):
LGTM!

PR #390, opened 2022-10-14
Allow doxygen generation to be conditional

ZehMatt:


athre0z (approved):
LGTM, thanks!

Tachi107:
Note: the documentation could be disabled already, by passing [`-DCMAKE_DISABLE_FIND_PACKAGE_Doxygen=true`](https://cmake.org/cmake/help/latest/variable/CMAKE_DISABLE_FIND_PACKAGE_PackageName.html)

athre0z:
Yeah, but the discoverability isn't great with that one. Most people aren't CMake experts and would likely never find this switch. With a dedicated option, it is immediately obvious how to disable it since it shows up in the CMake GUI (which I'd assume is probably what most people use, particularly on Windows).

Tachi107:
True. I became aware of that option just a few weeks ago, but I thought that it could be useful to know for you too :)

PR #391, opened 2022-10-15
Fixed ISA sets (Fixes #389)

### LINKED ISSUE #389: Some instructions have no `isa_set` assigned
`invlpgb,tlbsync` from extension `AMD_INVLPGB`
`psmash,pvalidate,rmpadjust,rmpupdate` from extension `SNP`

`isa_set` field is completely missing from JSON resulting in `ZYDIS_ISA_SET_INVALID` being reported. It seems that in case of all other lesser known extensions `isa_set == extension`, so I think same should apply here.

mappzor:
Although this isn't an important bug, fixing it causes enum change, so I thought it would be good to get it resolved before v4.

Fixes #389


flobernd (approved):
LGTM! Thanks.

flobernd:
Should we mention changes to the generated enums in the porting guide? Could be useful, but these will change a lot when new ISA sets/exts are added.

mappzor:
I was thinking about adding it but decided not to. I think it's expected that autogenerated files will change frequently between versions. Adding new instructions, ISA sets, etc. happens even with smaller releases.

Btw speaking of porting guide, I think it would be nice to mention `ZydisDisassemble*` API for discoverability.

flobernd:
Yeah I think the same. Probably overkill to mention the autogenerated stuff.

Regarding the "disassemble" API: I'm fine with mentioning it in the porting guide in addition to the changelogs. What do you think @athre0z ?

Anyways, going to merge this one right now.

PR #400, opened 2022-11-20
MSVC project updates

Mattiwatti:
This PR brings the project files under `/msvc` up to date with the current v4.0 tree. Most of the changes are fairly straightforward and the commit messages should be self-explanatory for these.

Quick TLDR summary of changes:
- MSVC projects now use VS2022, not VS2019. NB: this will likely break the CI build until the CI image is also updated to use VS2022 and the 10.0.22621.0 SDK/WDK.
- Added project files for the `Disassemble`, `DisassembleSimple` and `EncodeMov` samples.
- Support for building `ZydisWinKernel` for 32 bit targets has been dropped (not my decision, but the new WDK does not leave much room for choice). If anyone for some reason really wants to build this sample for 32 bit, then this can still be done using VS2019.
- Synced Zycore/Zydis header files used by the other projects in the solution. This does not affect the generated code, it only improves IntelliSense/Resharper context awareness and navigation when these headers are included.


flobernd (approved):
LGTM! Thanks a lot 🙂

PR #401, opened 2022-11-20
Disassembler: fix argument type `usize` -> `u64`

athre0z:
@Mattiwatti realized that the new disassembler functions have the wrong argument types -- this PR fixes that issue.


flobernd (approved):
Good catch!

PR #404, opened 2022-11-26
Optimize logo for darkmode and add logos to assets folder

oberrich:
GitHub has recently added an option to [control images based on light/dark color mode](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to) using the `<picture>` element.

This PR adds the relevant HTML to README and provides a dark/light mode logo under the `assets/img` folder. Shipping the logos adds around 6KB to the repo, load time should stay the same as the logo is currently also served through githubusercontent.com

PR #405, opened 2022-11-26
Use absolute URL for logo & make logo link to website

oberrich:
Fixes logo on mobile app (see https://github.com/zyantific/zydis/pull/404#issuecomment-1328056347)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 <p align="center">
   <picture>
-    <source media="(prefers-color-scheme: dark)" srcset="assets/img/logo-dark.svg" width="400px">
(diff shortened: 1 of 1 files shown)


flobernd (approved):
LGTM!

athre0z (approved):
Using this internal GitHub URL directly seems like something that has a pretty decent chance of randomly breaking at some point in the future. But honestly I don't really care: we can just revert to relative paths when it does. Personally I would also not have cared about breaking the iOS app since that is clearly a bug in the app.

flobernd:
Thinking the same here. That's why I did not merge right away. Should be reported as a bug for the iOS app.

PR #407, opened 2022-11-27
Update instructions and add new ISA extensions

flobernd:
## Improve several instruction definitions

- Change `LDTILECFG` and `STTILECFG` memory operand element type from `INT32` to `STRUCT`
- Require `VEX.L=0` for `JKZD/JKNZD`
- Set `FCOMI` isa-set for `F{U}COMI{P}`
- Change `MOV{H|L|LH|HL}P{S|D}` register access from `r` to `rw`
- Change `MASKMOVDQU` element types from `INT32` to `UINT8`
- Add implicit `GS` register operand to `SWAPGS`
- Change `RCPSS`, `ROUNDS{S|D}`, `SQRT{S|D}` and `RSQRTSS` operand access from `r` to `rw`

## Add new isa extensions

- `AMX-FP16`
- `AVX-IFMA`
- `AVX-NE-CONVERT`
- `AVX-VNNI-INT8`
- `RAO-INT`
- `IPREFETCH`
- `MSRLIST`
- `WRMSRNS`


athre0z (approved):
LGTM!

PR #418, opened 2023-02-16
CMake: remove `CXX` from `project` directive

athre0z:
Zydis currently requests a C++ compiler despite not actually using it. This breaks build on toolchains without a C++ compiler (e.g. musl libc).

@Tachi107 already did the same for zycore a while ago:

https://github.com/zyantific/zycore-c/commit/8f39333a9de5c868063587594f69d2c05de48c38


flobernd (approved):
LGTM!

PR #424, opened 2023-03-03
Use ZYAN_DIV64 in ZydisStringAppendDecU64

jpidancet:
Avoid a direct 64bit integer division in ZydisStringAppendDecU64 and use the macro provided by zycore instead, as GCC might generate unwanted calls to arithmetic functions on 32bit platforms.

This is the Zydis counterpart of zyantific/zycore-c#62

Note that this PR needs an update of the dependencies/zycore submodule first.

Thanks.


jpidancet:
Updated, thanks!

athre0z (approved):
LGTM, thanks!

PR #429, opened 2023-03-17
Add `PCOMMIT` instruction

### LINKED ISSUE #414: Support for PCOMMIT instruction.
Hi,

I am currently conducting an experiment using Zydis where we need to disassemble the linux kernel, including some rather old revisions of the kernel and I'm running into an issue where Zydis fails to recognize a particular instruction: PCOMMIT.

It turns out PCOMMIT is quite special since it was deprecated by Intel even before shipping CPUs supporting it (https://www.intel.com/content/www/us/en/developer/articles/technical/deprecate-pcommit-instruction.html). But it is a problem when disassembling older distro kernels since there are still linux kernels in the wild that have been compiled with this instruction and we need to be able to disassemble it properly.

If adding support for this instruction in Zydis's mainline tree is not possible, we would be grateful to have the instructions on how to manually add support for custom instructions to our own tree. As I understand, currently a lot of code from the Zydis source is actually generated from files in the zydis-db project, but I was unable to find proper documentation on how to use it.

Thanks.

flobernd:
Closes #414


athre0z (approved):
👍

jpidancet:
Many thanks.

PR #430, opened 2023-03-20
Explicitly zero ZydisDisassemble instruction

chriso:
Firstly, thank you for the brilliant library! 

I ran into an issue while compiling the library to webassembly using clang (`-target wasm32-freestanding`).

The `-DZYAN_NO_LIBC` flag works as expected, but I noticed that clang inserts a `memset` call when the `ZydisDisassembledInstruction *instruction` argument of `ZydisDisassemble` is [implicitly zeroed](https://github.com/zyantific/zydis/blob/aeb1b9acf0703fd56dae879c99b05704ea955511/src/Disassembler.c#L43-L46). If I explicitly zero the instruction with `ZYAN_MEMSET` (this PR) then I get the expected result; no `memset` call there, and no libc calls anywhere.

clang doesn't seem to respect the `-ffreestanding` and `-fno-builtins` flags. I'm not sure what flag I need to use (or if one exists) to eliminate the `memset` call. I appreciate that this is probably a clang issue rather than something zydis should concern itself with, but I'll submit this PR either way in case others run into the same issue in future.

### DIFF
--- a/src/Disassembler.c
+++ b/src/Disassembler.c
@@ -40,10 +40,8 @@ static ZyanStatus ZydisDisassemble(ZydisMachineMode machine_mode,
         return ZYAN_STATUS_INVALID_ARGUMENT;
     }
 
-    *instruction = (ZydisDisassembledInstruction)
-    {
-      .runtime_address = runtime_address
-    };
+    ZYAN_MEMSET(instruction, 0, sizeof(*instruction));
+    instruction->runtime_address = runtime_address;
 
     // Derive the stack width from the address width.
     ZydisStackWidth stack_width;


flobernd:
Hi there, thanks for your contribution! 

I currently have no time to do in depth profiling with multiple compilers, but I assume both pieces of code will probably result in the same instructions/memset call (in regular LIBC mode) as the ZydisDecodedInstruction struct is rather large. In the worst case we could conditionally use ZYAN_MEMSET only for NO_LIBC mode (or even NO_LIBC+CLANG).

Maybe @athre0z has time to check this..?

Anyways I think this place should use ZYDIS_MEMSET in all cases to give users the option to plug in their own LibC replacement functions for improved performance. We don't have this option, if we rely on the compiler generated code for struct initialization (although this is probably the more optimized version in general, compared to our default LiC replacement functions).

chriso:
Thanks for the quick replies @flobernd and @athre0z!

You're right, clang detects `memset` and `memcpy` like functions/idioms and replaces them with libc calls. From what I can gather you can disable this behavior with the `-disable-simplify-libcalls` flag, but it's only accepted by the lower-level LLVM `opt` and `llc` tools. I can't find a way to thread this flag through clang to LLVM. There's more info here: https://stackoverflow.com/questions/21318229/llvm-optimizes-with-library-functions.

When I compile with `-O0 -target wasm32-freestanding -DZYAN_NO_LIBC` I see that there are both `memset` and `memcpy` calls present in the resulting WASM module. If I bump the optimization level to either  `-O1`, `-Os` or `-Oz`, the calls all disappear except for the one `memset` call that this PR addresses. I get the same results at all other optimization levels. My guess is that the sizes are small and known in all other cases, and so it's trivial for LLVM to replace them with inline loads/stores. If we look at this one `memset` instance, it appears that the struct is large (`sizeof(ZydisDisassembledInstruction) == 1232`) and so probably too large to inline the stores?

It doesn't sound like it's possible to solve this issue in the general case, however if we explicitly zero memory at this one site then it becomes possible to compile the library to WASM without any libc calls (at least with clang v15).

flobernd:
If it does not hurt performance, I'm in favor of accepting your PR 🙂 

One way to reliably solve the problem for all optimization levels on your side, would be to define your own memset and memcpy functions somewhere in your code and use the amalgamated version of Zydis. This should cause clang to call your custom code instead of the original libc functions. At least that's how I interpreted @athre0z comment.

athre0z:
> If it does not hurt performance, I'm in favor of accepting your PR slightly_smiling_face

IMHO it sets a bad precedent. We have a lot of places in Zydis where we initialize relatively large structs and I don't really think that it is practical for us to constantly keep up with optimization passes and size thresholds before library calls are emitted. This will likely also depend on `-mtune` and `-march` as well as the target tuple. We'll essentially have to stop using struct initializer literals completely again if we want to make sure this is never generated. 

If we really want to take the burden of dealing with what I personally consider a bug in clang, a much more reliable approach to do so would be to offer e.g. a macro `ZYAN_DEFINE_CLANG_BUILTINS` that simply defines and exports `memset`, `memcpy` and `memmov` functions for clang to use which simply call into our `ZYAN_XXX` functions.

flobernd:
Mhh, true I forgot about the other places that could potentially lead to emitting calls to the LibC builtins. A macro/option sounds good.

athre0z:
> Edit: We could even do this implicitly without a macro, if we detect CLang and ZYAN_NO_LIBC is enabled.

I wouldn't go that far. Most no-libc environments will already define these symbols and our defines would then result in clashes.

PR #435, opened 2023-04-18
Convert ZydisDecoder field decoder_mode to a bitmap.

jpidancet:
Make decoder_mode a bitmap instead of an array of booleans for space efficiency.

### DIFF
--- a/src/Decoder.c
+++ b/src/Decoder.c
@@ -2103,7 +2103,7 @@ static void ZydisSetAttributes(ZydisDecoderState* state, ZydisDecodedInstruction
                     break;
                 }
             }
-            if (state->decoder->decoder_mode[ZYDIS_DECODER_MODE_MPX] &&
+            if ((state->decoder->decoder_mode & (1 << ZYDIS_DECODER_MODE_MPX)) &&
                 instruction->attributes & ZYDIS_ATTRIB_ACCEPTS_BND)
             {
                 instruction->attributes |= ZYDIS_ATTRIB_HAS_BND;
@@ -2167,7 +2167,7 @@ static void ZydisSetAttributes(ZydisDecoderState* state, ZydisDecodedInstruction
         if (def->accepts_NOTRACK)
         {
             instruction->attributes |= ZYDIS_ATTRIB_ACCEPTS_NOTRACK;
-            if (state->decoder->decoder_mode[ZYDIS_DECODER_MODE_CET] &&
+            if ((state->decoder->decoder_mode & (1 << ZYDIS_DECODER_MODE_CET)) &&
                 (state->prefixes.offset_notrack >= 0))
             {
                 instruction->attributes |= ZYDIS_ATTRIB_HAS_NOTRACK;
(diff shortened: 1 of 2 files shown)


src/Decoder.c
```diff
@@ -4899,23 +4899,15 @@ static ZyanStatus ZydisDecodeInstruction(ZydisDecoderState* state,
...
-        ZYAN_FALSE, // ZYDIS_DECODER_MODE_KNC
-        ZYAN_TRUE , // ZYDIS_DECODER_MODE_MPX
-        ZYAN_TRUE , // ZYDIS_DECODER_MODE_CET
-        ZYAN_TRUE , // ZYDIS_DECODER_MODE_LZCNT
-        ZYAN_TRUE , // ZYDIS_DECODER_MODE_TZCNT
-        ZYAN_FALSE, // ZYDIS_DECODER_MODE_WBNOINVD
-        ZYAN_TRUE,  // ZYDIS_DECODER_MODE_CLDEMOTE
-        ZYAN_FALSE  // ZYDIS_DECODER_MODE_IPREFETCH
-    };
+        (1 << ZYDIS_DECODER_MODE_MPX) ||
```
flobernd, line 4906: I think these should use `|` instead of `||`

src/Decoder.c
```diff
@@ -4777,31 +4777,31 @@ static ZyanStatus ZydisDecodeInstruction(ZydisDecoderState* state,
             break;
 #endif
         case ZYDIS_NODETYPE_FILTER_MODE_AMD:
-            index = state->decoder->decoder_mode[ZYDIS_DECODER_MODE_AMD_BRANCHES] ? 1 : 0;
+            index = !!(state->decoder->decoder_mode & (1 << ZYDIS_DECODER_MODE_AMD_BRANCHES));
```
flobernd, line 4780: Seems like some compilers on some platforms are not happy with this 😛

flobernd (changes requested):
I like this change, but it seems like some stuff needs to be fixed before merging 🙂 

The regression tests are failing and the fuzzer complains as well.

src/Decoder.c
jpidancet: Oops.

jpidancet:
Hi,

Thank you for having a look. It looks like the `||` vs `|`  mix-up was causing all the build and fuzzing errors. It is fixed now.

flobernd (approved):

athre0z:
> For v5 I would like to change the public API for this entirely (ZydisDecoderEnableMode(mode) -> ZydisDecoderSetModes(bitmask) and converting the ZYDIS_DECODER_MODE_X constants to bits), but for now it's probably better to keep the current API.

We can do this in a non-breaking manner by just adding such a function without removing the old one. I think this would be preferable here since maintaining the old function is very little effort. We can mark the old function as deprecated and remove it in a few years, or just keep it around forever. Changing the constants is not API breaking, only ABI breaking, so can be done in a minor version.

athre0z (approved):
LGTM, good change!

PR #437, opened 2023-04-30
Add ZYDIS_BUILD_TESTS option and include tests in CMakeLists.txt

topazus:


athre0z:
Thanks, that's good to have! I added two more commits to also rework our convenience Makefile and the CI to go via the CMake testing path as well now for the sake of de-duplication.

athre0z:
Ok, it's also working on Windows now. I had to replace the literal `python3` with the variable populated by `find_package(Python ...)`, add generator expressions for the executable paths (on Windows the executables are placed into sub-directories of the output directory) and add `if (TARGET xyz)` guards to make sure it doesn't break when e.g. the encoder is disabled.

athre0z (approved):

topazus:
Thanks for the review and further improvement.

PR #440, opened 2023-05-20
Build: fix amalgamation public header warning

athre0z:
The previous code would mix up absolute and relative paths to determine when to print the warning about public headers first being encountered when processing the source files. This commit fixes this and thus gets rid of the incorrect warnings.


flobernd (approved):
LGTM!

PR #441, opened 2023-05-21
Fix the installation location of manpages

topazus:
Hi, I am trying to make zycore-c and zydis into Fedora official repository. This PR is to change the install location of manpages of small useful CLI tools to `${CMAKE_INSTALL_MANDIR}/man1` to accomodate the common rules, which the tools will be built with the `ZYDIS_BUILD_TOOLS` option is on.


athre0z (approved):
Thanks, LGTM!

PR #442, opened 2023-05-21
Add missing `(void)` on functions without args

athre0z:
Multiple functions without arguments were missing the `(void)` that prevents them from being interpreted as not having a prototype. Newer Clang versions started printing warnings for that in pedantic mode.

Corresponding Zycore PR: https://github.com/zyantific/zycore-c/pull/64/files


flobernd (approved):
LGTM!

PR #450, opened 2023-09-12
Add `AMX` ISA extension

flobernd:


src/Segment.c
```diff
@@ -88,6 +88,12 @@ ZyanStatus ZydisGetInstructionSegments(const ZydisDecodedInstruction* instructio
         segment_size = 4;
         break;
     default:
+        if (instruction->attributes & ZYDIS_ATTRIB_HAS_REX2)
```
flobernd, line 91: This should check encoding instead.

tools/ZydisInfo.c
```diff
@@ -777,6 +810,10 @@ static void PrintAVXInfo(const ZydisDecodedInstruction* instruction)
             ZydisRegisterGetString(instruction->avx.mask.reg),
             CVT100_OUT(COLOR_VALUE_LABEL), CVT100_OUT(COLOR_VALUE_B),
             strings_mask_mode[instruction->avx.mask.mode], CVT100_OUT(COLOR_VALUE_LABEL));
+        if (instruction->attributes & ZYDIS_ATTRIB_HAS_SCC)
```
flobernd, line 813: We should probably put this in a dedicated APX section together with the "zu" (which is not implemented yet at all).

src/Decoder.c
```diff
@@ -2351,6 +2402,8 @@ static void ZydisSetAVXInformation(ZydisDecoderContext* context,
         const ZydisInstructionDefinitionEVEX* def =
             (const ZydisInstructionDefinitionEVEX*)definition;
 
+        // TODO: Don't set vector length, etc. for scalar instructions
```
flobernd, line 2405: This requires a dedicated flag and generator support.

src/Decoder.c
```diff
@@ -3792,6 +3922,39 @@ static ZyanStatus ZydisNodeHandlerOpcode(ZydisDecoderState* state,
         default:
             ZYAN_UNREACHABLE;
         }
+
+        if (instruction->encoding != ZYDIS_INSTRUCTION_ENCODING_REX2)
+        {
+            break;
+        }
+
+        ZyanBool illegal_rex2 = (instruction->opcode == 0x0F);
```
flobernd, line 3931: Might be better to just generate a flag for this in the generator based on the same condition. We usually try to not hardcode table related conditions.

src/Decoder.c
```diff
@@ -3445,6 +3499,16 @@ static void ZydisSetEffectiveOperandWidth(ZydisDecoderContext* context,
     };
 
     ZyanU8 index = (instruction->attributes & ZYDIS_ATTRIB_HAS_OPERANDSIZE) ? 1 : 0;
+
+    if ((instruction->encoding == ZYDIS_INSTRUCTION_ENCODING_EVEX) &&
+        (instruction->opcode_map == ZYDIS_OPCODE_MAP_MAP4) &&
+        (instruction->raw.evex.pp == 0x01))
+    {
+        // EVEX encoded instructions in MAP4 must use 0x66 as the mandatory prefix AND the
```
flobernd, line 3507: Might be better to have a flag for this. We usually don't hardcode table related conditions.

src/Decoder.c
```diff
@@ -4113,6 +4282,82 @@ static ZyanStatus ZydisNodeHandlerEvexB(const ZydisDecodedInstruction* instructi
...
+
+    // Error conditions
+
+    if (instruction->raw.evex.z || context->vector_unified.L || 
+        (context->vector_unified.mask & 0x03))
+    {
+        return ZYDIS_STATUS_DECODING_ERROR;
+    }
+
+    // TODO: NF without ND filter -> evex.b = zero upper
```
flobernd, line 4322: Generator flag would be better here.

src/Decoder.c
```diff
@@ -4113,6 +4282,82 @@ static ZyanStatus ZydisNodeHandlerEvexB(const ZydisDecodedInstruction* instructi
...
+    }
+
+    // APX conditional CMP and TEST uses a special form of the EVEX prefix which reuses
+    // the `.V4` bit and the `.aaa` bits as the `.SCC` condition code and the `.vvvv` bits
+    // as the DFV specifier. Other than the regular `.vvvv` NDS/NDD specifier, the DFV
+    // specifier bits are not inverted.
+
+    // All APX conditional CMP and TEST instructions must have a SCC filter!
+
+    context->vector_unified.vvvv = (~context->vector_unified.vvvv) & 0x0F;
```
flobernd, line 4352: This is clumsy.. It would be nice if the generator emits a hint about the EVEX variant to use. We are currently just "guessing".

src/Decoder.c
```diff
@@ -4172,16 +4417,34 @@ static ZyanStatus ZydisPopulateRegisterIds(ZydisDecoderContext* context,
         const ZyanBool is_emvex = (instruction->encoding == ZYDIS_INSTRUCTION_ENCODING_EVEX) ||
```
flobernd, line 4417: This function needs to be revisited. For CR and DR the high 16 register are invalid. For other types they are ignored.

src/Decoder.c
flobernd: There are as well XSAVE and XRSTORE variants that do not support REX2.

src/Decoder.c
flobernd: EVEX.b is actually EVEX.ND in these cases.

src/Decoder.c
```diff
@@ -1009,6 +1053,7 @@ static void ZydisSetOperandSizeAndElementInfo(const ZydisDecoderContext* context
 #ifndef ZYDIS_DISABLE_AVX512
             if (definition->size[context->eosz_index])
             {
+                // TODO: fixme
```
flobernd, line 1056: This block is not entered for some tile instructions, resulting in a panic.

src/Register.c
```diff
@@ -202,7 +202,7 @@ ZydisRegister ZydisRegisterGetLargestEnclosing(ZydisMachineMode mode, ZydisRegis
         return static_reg;
     }
 
-    static const ZyanU8 GPR8_MAPPING[20] =
+    static const ZyanU8 GPR8_MAPPING[36] =
```
flobernd, line 205: We should change semantics of all `Register*` APIs based on a `ZyanBool apx` parameter.

mappzor (approved):

PR #453, opened 2023-09-13
Improve fuzzer stability in persistent mode

athre0z:
`fread` does internal buffering that can result in different code paths to be taken every time we call it. This is detrimental for fuzzing stability in persistent mode. This PR changes the code to use a direct `read` syscall on POSIX.


flobernd (approved):
LGTM!

PR #460, opened 2023-10-15
Encoder: rename `operand_count` -> `operand_count_visible`

### LINKED ISSUE #458: Question about ZydisEncoderDecodedInstructionToEncoderRequest's operand_count parameter
This parameter is called `operand_count `, however, there is a runtime check at https://github.com/zyantific/zydis/blob/master/src/Encoder.c#L4639 that requires it to be less than or equal to `instruction->operand_count_visible`. So maybe it should be renamed to `operand_count_visible` to reduce confusion?

An example is `shl rcx, 0x07`. This would fail if using `ZydisEncoderDecodedInstructionToEncoderRequest(
    &insn.info, insn.operands, insn.info.operand_count, &req);`

athre0z:
Resolves #458

CC @mappzor

### DIFF
--- a/assets/porting-guide-v4-v5.md
+++ b/assets/porting-guide-v4-v5.md
@@ -1,0 +1,7 @@
+# Porting Guide v4 -> v5
+
+# Encoder
+
+- `ZydisEncoderDecodedInstructionToEncoderRequest` now expects exactly `instruction->operand_count_visible` to be
+  passed, not `operand_count_visible` at maximum. Passing a lower value was previously allowed but didn't really
+  make much sense at all.
--- a/src/Encoder.c
+++ b/src/Encoder.c
@@ -4525,10 +4525,11 @@ ZYDIS_EXPORT ZyanStatus ZydisEncoderEncodeInstructionAbsolute(ZydisEncoderReques
 }
 
 ZYDIS_EXPORT ZyanStatus ZydisEncoderDecodedInstructionToEncoderRequest(
-    const ZydisDecodedInstruction *instruction, const ZydisDecodedOperand* operands, 
-    ZyanU8 operand_count, ZydisEncoderRequest *request)
+        const ZydisDecodedInstruction *instruction, const ZydisDecodedOperand* operands,
+        ZyanU8 operand_count_visible, ZydisEncoderRequest *request)
 {
-    if (!instruction || !request || (operand_count && !operands))
+    if (!instruction || !request || (operand_count_visible && !operands) ||
+        operand_count_visible != instruction->operand_count_visible)
     {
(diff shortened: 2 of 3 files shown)


src/Encoder.c
```diff
@@ -4635,13 +4635,13 @@ ZYDIS_EXPORT ZyanStatus ZydisEncoderDecodedInstructionToEncoderRequest(
     }
     request->allowed_encodings = 1 << instruction->encoding;
 
-    if ((operand_count > ZYDIS_ENCODER_MAX_OPERANDS) || 
-        (operand_count > instruction->operand_count_visible))
+    if ((operand_count_visible > ZYDIS_ENCODER_MAX_OPERANDS) ||
+        (operand_count_visible > instruction->operand_count_visible))
```
athre0z, line 4639: @mappzor wondering whether this should really be `==`. Does it ever make sense to pass a lower number? It's an API breaking chance, so I didn't apply it, but might make sense to change this in v5?

mappzor: Those checks were changed here: https://github.com/zyantific/zydis/commit/875c8a75344d0ad71b02bcd2cc5e3acb87e7b003#diff-b27231e6044d9f536c52c09cb128aee56c82926c92676b480bd438df1f96787aL4407

It was surely the safest change to make considering how old code used to work but I don't think it ever makes sense to pass a lower number. Even if there was such case I'd consider it a bad way to use the API. It's assumed that result of decoding is passed without changes to `ZydisEncoderDecodedInstructionToEncoderRequest` and eventual adjustments to operand count, types, etc. are applied manually to resulting request. I'm in favor of `==` check in future version.

flobernd: I think my PR with the ZydisRegisterGetLargestEnclosing was already breaking and APX definitely is, so I'd say "go for it"!

We might have to release a new major version soon anyways.

flobernd (approved):

src/Encoder.c
athre0z: Yeah. The `ZydisRegisterGetLargestEnclosing` was indeed API breaking, though in a similar way as this one: you'd have to use the function in a really weird way to realize it. But that doesn't change that it was API breaking.

I adjusted the code to check with `==`. I further moved the check up and combined it with the other trivial argument validation.

flobernd (approved):

assets/porting-guide-v4-v5.md
```diff
@@ -0,0 +1,7 @@
+# Porting Guide v4 -> v5
```
flobernd, line 1: I should add the ZydisRegisterGetLargestEnclosing change here as well 😋

PR #461, opened 2023-10-15
CMake: improve search for Zycore

athre0z:
The changes in this PR should allow Zydis to find Zycore in more situations automatically. Most importantly, `find_package(Zycore)` is now something that we try automatically. This should allow distributions that ship Zycore as a separate package to do so more easily without having to delve into out CMake logic to discover the `ZYAN_SYSTEM_ZYCORE` option.

I tested:

- building a full clone with submodule
- building an incomplete clone (with and without git installed)
- fallback to system zycore (by hacking the NixOS package, with and without `-DZYAN_SYSTEM_ZYCORE`)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,7 @@ option(ZYDIS_LIBFUZZER
 
 # Dependencies
 option(ZYAN_SYSTEM_ZYCORE
-    "Use system Zycore library"
+    "Force using system installed Zycore library"
     OFF)
 set(ZYAN_ZYCORE_PATH
     "${CMAKE_CURRENT_LIST_DIR}/dependencies/zycore"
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -83,35 +83,62 @@ set(ZYAN_ZYCORE_PATH
...
+    if (Zycore_FOUND)
+        message(VERBOSE "Using system Zycore")
+        return ()
+    endif ()
+
+    if (ZYAN_SYSTEM_ZYCORE)
+        message(FATAL_ERROR "ZYAN_SYSTEM_ZYCORE set but no system-installed Zycore found")
+    endif ()
+
+    find_package(Git QUIET)
```
flobernd, line 119: We could possibly change this block to the newer CMake 'FetchContent' APIs. However, not sure if we would have to bump our CMake version requirement to do so.

flobernd (approved):
LGTM! Good improvement.

CMakeLists.txt
athre0z: I suspect we'd have to maintain a separate pinned revision in the CMakeLists then, no? With the existing approach we're using the revision stored in git's meta-info. Or does `FetchContent` explicitly support this somehow?

flobernd: I think we would have to either maintain it in two places or extract the hash from the repository config by parsing the Git files.

Not sure if it's worth it. Probably not. Git should be installed on most dev machines anyways.

PR #464, opened 2023-11-26
Fix #463: Wrong target offset calculated when using bnd prefix

### LINKED ISSUE #463: Encoder generates wrong relative offset with bnd prefix
`[ERR] Assertion failed: bnd jnb 0x7FF64AEE5472 == bnd jnb 0x7FF64AEE5473`

Found this one while adding rewriter validation to my instrumentation tool that uses Zydis, the offset is off by 1 when it contains `ZYDIS_ATTRIB_HAS_BND`

It seems that ZydisTestEncoderAbsolute doesn't cover this case, adding the prefix makes it fail the tests.

ZehMatt:


tools/ZydisTestEncoderAbsolute.c
```diff
@@ -224,6 +224,7 @@ static ZyanBool RunBranchingTests(void)
     static const ZydisInstructionAttributes prefixes[] = {
         0,
         ZYDIS_ATTRIB_HAS_BRANCH_TAKEN,
+        ZYDIS_ATTRIB_HAS_BND,
     };
     static const char *str_prefixes[] =
```
mappzor, line 229: This table has to be updated every time `prefixes` is updated. Suggested name: `PBN` or `PBD`.

ZehMatt: Ah darn I missed that, I think PBN is a good choice.

src/Encoder.c
```diff
@@ -4373,7 +4373,8 @@ ZYDIS_EXPORT ZyanStatus ZydisEncoderEncodeInstructionAbsolute(ZydisEncoderReques
                     }
                     if ((rel_info->accepts_branch_hints) &&
```
mappzor, line 4374: We cannot depend on just `accepts_branch_hints` here and in tests. It's only for `BRANCH_TAKEN` and `BRANCH_NOT_TAKEN`. This means `BND` fix covers BND+JCC but doesn't cover `BND+JMP` which doesn't accept branch hints.

mappzor: BND+CALL is affected as well.

ZehMatt:
I believe this PR is good to go, @mappzor confirmed on Discord that nothing else needs adjusting.

athre0z (approved):
Thanks you two!

PR #466, opened 2023-12-09
Fuzzing improvements and bugfixes

mappzor:
I'm putting 3 changes in this PR because fuzzers naturally depend on bugfixes and those are one-liners anyway. Having separate pull requests would be an overkill and they would have to be merged in a specific order. I've kept one change per commit for clarity.

Changes:
- Fuzzing `ZydisEncoderEncodeInstructionAbsolute` - primary goal of this PR, doesn't need much explanation. Although `ZydisTestEncoderAbsolute` aims to cover every case some stuff went unnoticed like #463. Fuzzing is a valuable 2nd layer of testing here.
- 3DNow! handling - first and hopefully the last issue found by new fuzzing logic. 3DNow! instructions have extra opcode byte at the end and that confused logic inside `ZydisEncoderEncodeInstructionAbsolute`.
- Enable IPREFETCH mode by default - we had it documented as enabled by default while it wasn't


flobernd (approved):
LGTM! Thank you!

athre0z (approved):
LGTM as well!

PR #467, opened 2023-12-09
Doc: extract documentation from all headers 

### LINKED ISSUE #465: Missing enum definitions in Documentation
[Here](https://doc.zydis.re/v4.0.0-rc.1/html/structZydisDecodedInstructionMeta__) we can see the documentation for meta information of decoded instruction with fields which are not available directly. It makes users forced to clone the repo from github and manually grepping the needed enum definitions. 

For example,  **ZydisInstructionCategory** is defined in **include/Zydis/Generated/EnumInstructionCategory.h** and is invisible in documentation.

athre0z:
We previously went with explicitly adding `@file` directives to our headers instead of using `EXTRACT_ALL` because the latter was broken with the m.css theme that we were using previously. m.css has a custom code generator that didn't support the directive. We've since switched to another theme that just re-styles the vanilla HTML, so we're fine to just use that now (instead of changing the code-gen to add the directives in all generated headers).

Also: update Doxygen CSS theme to latest.

Resolves #465


flobernd (approved):
LGTM!

PR #472, opened 2024-01-11
Formatter improvements (sizes and addresses)

mappzor:
1. Replaced `stack_width` -> `address_width` to prevent trimming real values in some situation (address size overrides)
1. Print `agen` sizes (in forced mode only) -> affects `lea` and some `bnd*` instructions, `mib` behavior unchanged because we don't have standardized names for such sizes anyway


flobernd (approved):
LGTM! Thank you.

athre0z (approved):
LGTM as well!

PR #473, opened 2024-01-12
Reworked absolute address handling (Fixes #471)

### LINKED ISSUE #471: Encoder mishandles 16-bit address truncation behavior
Let's consider two instructions in 32-bit mode:

```
00 3d 00 ff ff ff       add byte ptr ds:[0xFFFFFF00], bh
67 00 3e 00 ff          add byte ptr ds:[0x0000FF00], bh
```

Attempting to re-encode both of them results in `67 00 3e 00 ff add byte ptr ds:[0x0000FF00], bh`. This presents a unique problem as it's not just a matter of buggy logic. In both cases disassembled instructions have matching operands (fuzzer couldn't catch this) with full 64-bit displacement `0xFFFFFFFFFFFFFF00`. The only difference is effective address size (attribute of the whole instruction not just a single operand), so it's all about interpreting the operand correctly.

I see two ways of fixing this:
1. require 16-bit absolute addresses to be truncated to 16-bits by the user - I don't like this, it's very clunky and goes against the rule that displacements are uniformly signed, it's also just a terrible API
1. use address size hints - `ZydisEncoderDecodedInstructionToEncoderRequest` always sets those hints anyway based on effective address size, currently they are ignored unless instruction scales with address size (hidden scaling). This hint can be used in additional context to resolve disambiguity between 16 and 32-bit absolute addresses. I think it's intuitive that address size hint covers all issues regarding address size. It's also a very natural place to document this properly. Lack of hint would indicate that native size is preferred (16 in 16-bit modes, 32 bits otherwise)

Currently I'm exploring 2nd option as it seems the most viable. I'm also considering to get rid of another clunky 16-bit behavior. Consider this in 16-bit mode:
```
67 3A 05 00 1E 57 1A    cmp al, byte ptr ds:[0x1E00]
```

Displacement is `0x1A571E00` after disassembling but it gets truncated to 16 bits inside `ZydisCalcAbsoluteAddress` during formatting. Currently encoder ignores everything above 16 bits to re-encode this properly. I think encoder shouldn't accept this and it would be better to make `ZydisEncoderDecodedInstructionToEncoderRequest` responsible for sanitization.

mappzor:
This PR fixes all known issues around treating signed displacements as absolute addresses. Address size hint is now used for disambiguation where needed (rare cases that essentially force address override prefix). This also lead to a simplification of the fuzzer as verification of memory operands was always quite messy due to multiple issues that it had to deal with (signedness, 16-bit truncation, MIB, etc.). Now it's finally replaced with simpler and more robust code.

Fixes #471

### DIFF
--- a/src/Encoder.c
+++ b/src/Encoder.c
@@ -1578,17 +1578,6 @@ static ZyanBool ZydisIsMemoryOperandCompatible(ZydisEncoderInstructionMatch *mat
             {
                 return ZYAN_FALSE;
             }
-            if (ZydisGetMachineModeWidth(match->request->machine_mode) == 16)
-            {
-                if ((ZyanI16)displacement == 0)
-                {
-                    disp_size = 0;
-                }
-                else
-                {
-                    disp_size = ZydisGetSignedImmSize((ZyanI16)displacement);
-                }
-            }
 
             match->cd8_scale = ZydisGetCompDispScale(match);
             if (match->cd8_scale)
@@ -1596,7 +1585,8 @@ static ZyanBool ZydisIsMemoryOperandCompatible(ZydisEncoderInstructionMatch *mat
                 const ZyanI64 mask = (1 << match->cd8_scale) - 1;
                 if (!(displacement & mask))
                 {
-                    disp_size = ZydisGetSignedImmSize(displacement >> match->cd8_scale);
+                    if (ZydisGetSignedImmSize(displacement >> match->cd8_scale) == 8)
+                        disp_size = 8;
                 }
                 else if (disp_size == 8)
                 {
@@ -1941,7 +1931,7 @@ static ZyanBool ZydisIsMemoryOperandCompatible(ZydisEncoderInstructionMatch *mat
                     reg_index_class);
             }
         }
-        else
+        else if (disp_size != 8 || !match->cd8_scale)
         {
             ZyanU8 min_disp_size = match->easz ? match->easz : 16;
             if (((min_disp_size == 16) && !(match->definition->address_sizes & ZYDIS_WIDTH_16)) ||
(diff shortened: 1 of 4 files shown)


src/Encoder.c
```diff
@@ -1578,25 +1578,15 @@ static ZyanBool ZydisIsMemoryOperandCompatible(ZydisEncoderInstructionMatch *mat
...
 
             match->cd8_scale = ZydisGetCompDispScale(match);
             if (match->cd8_scale)
             {
                 const ZyanI64 mask = (1 << match->cd8_scale) - 1;
                 if (!(displacement & mask))
                 {
-                    disp_size = ZydisGetSignedImmSize(displacement >> match->cd8_scale);
+                    if (ZydisGetSignedImmSize(displacement >> match->cd8_scale) == 8)
+                        disp_size = 8;
```
flobernd, line 1589: Don't feel bothered to change this right now; just noticed the parenthesises are missing here. I think we should auto-format the complete codebase anyways at some point and also get rid of the strange `else` indentation I've been using in the rest of the code.

mappzor:
In the latest set of changes I've finally dealt with the source of all address-related problems. It's this piece of code from `ZydisEncoderDecodedInstructionToEncoderRequest`:
```c
if (dec_op->encoding == ZYDIS_OPERAND_ENCODING_DISP16_32_64)
{
	ZydisCalcAbsoluteAddress(instruction, dec_op, 0,
		(ZyanU64 *)&enc_op->mem.displacement);
}
else
{
	enc_op->mem.displacement = dec_op->mem.disp.has_displacement ?
		dec_op->mem.disp.value : 0;
}
```

It opened the gates of hell because it introduces special treatment for absolute addresses coming from `moffs` variants of `mov` instructions. So `mov [address], eax` could result in different encoder requests depending on which variant of mov it was ("normal" vs `moffs`). Dealing with this on the other end of the story (the encoder itself) resulted in convoluted processing of displacement-only and `moffs` operands. At some later point 16-bit workarounds were added to the mix.

Now things are done The Right Way :tm:. All displacements are the same and they are handled just like in other parts of Zydis (decoder and utils like `ZydisCalcAbsoluteAddress`) i.e. they are 64-bit signed and interpreted in the presence of external size information. For encoder this information is now the address size hint.

This took way more time and effort than I've originally anticipated for something that initially seemed like a minor oversight but I think the changes are worth it. New behaviors are truly uniform and shield the user from all the ugly quirks of `mov`. Handling of `none` hint should be more aligned with the expectations of a typical user as well (when encoding from scratch).

flobernd (approved):
LGTM! This looks way better from a usability perspective 🙂 Thanks a lot!

athre0z (approved):
LGTM from a coarse glance! Apologies for the delay: I didn't find the time to review this properly. Realistically you know this piece of code better than both Florian and me anyway, so let's not delay this any further. :)

PR #477, opened 2024-01-22
Add `UD0_COMPAT` decoder mode

### LINKED ISSUE #436: UD0 instruction without ModRM byte
Hi,

The UD0 instruction is a bit special in that it can appear with or without a ModR/M byte:

In Intel® 64 and IA-32 Architectures Software Developer’s Manual ( _INSTRUCTION SET REFERENCE, M-U_), it is described **with** a ModR/M byte,

However, it also says this:

> UD - Undefined Instruction:
> [...]
> NOTES:
>  1. Some processors decode the UD0 instruction without a
>     ModR/M byte. As a result, those processors would deliver
>     an invalid-opcode exception instead of a fault on
>     instruction fetch when the instruction with a
>     ModR/M byte (and any implied bytes) would cross a page
>     or segment boundary.
> 

Now, my problem is that it appears **without** the ModRM byte in some old versions of the Linux kernel: https://elixir.bootlin.com/linux/v4.14.313/source/arch/x86/include/asm/refcount.h#L17, so using Zydis to disassemble code from these kernels will fail because it will attempt to consume 1 byte too many.

I agree this is a difficult one to solve since there's no way for Zydis to evaluate wether the instruction length is 2 or 3 since it depends on the context. For info, capstone seems to ignore the ModR/M byte and views it as a 2byte only instruction.

Is there any way we could add a feature flag to control whether Zydis should accept or reject the ModR/M byte for this instruction ?

flobernd:
Closes #436


jpidancet:
Hi @flobernd, thanks a lot for working on this.
I verified you changes give the expected results using this simple program:

```
#include <Zydis.h>
#include <assert.h>

int main(void)
{
	ZydisDecoder decoder;
	const char data[4] = "\x0f\xff\x00\x00";
	ZydisDecodedInstruction instruction;
	ZydisDecodedOperand operands[ZYDIS_MAX_OPERAND_COUNT];
	ZyanStatus status;

	ZydisDecoderInit(&decoder, ZYDIS_MACHINE_MODE_LONG_64,
			 ZYDIS_STACK_WIDTH_64);
	ZydisDecoderEnableMode(&decoder, ZYDIS_DECODER_MODE_UD0_COMPAT, ZYAN_FALSE);
	status = ZydisDecoderDecodeFull(&decoder,
					data, sizeof(data),
					&instruction,
					operands);
	assert(ZYAN_SUCCESS(status));
	assert(instruction.mnemonic == ZYDIS_MNEMONIC_UD0);
	assert(instruction.length == 3);

	ZydisDecoderEnableMode(&decoder, ZYDIS_DECODER_MODE_UD0_COMPAT, ZYAN_TRUE);
	status = ZydisDecoderDecodeFull(&decoder,
					data, sizeof(data),
					&instruction,
					operands);

	assert(ZYAN_SUCCESS(status));
	assert(instruction.mnemonic == ZYDIS_MNEMONIC_UD0);
	assert(instruction.length == 2);

	return 0;	
}
```

flobernd:
I think the encoder does not like this new mode. Maybe @mappzor can have a look 🙂

mappzor:
Encoder itself is fine, it can encode everything based on mnemonics, signatures and encodings. `ZydisFuzzEncoder` starts with a random encoder request, so it needs to figure out how to configure decoder for verification purposes.

I would just add this piece of code [here](https://github.com/zyantific/zydis/blob/ud0-compat/tools/ZydisFuzzEncoder.c#L289):
```c
    if (request.mnemonic == ZYDIS_MNEMONIC_UD0 && request.operand_count == 0)
    {
        status = ZydisDecoderEnableMode(&decoder, ZYDIS_DECODER_MODE_UD0_COMPAT, ZYAN_TRUE);
        if (!ZYAN_SUCCESS(status))
        {
            fputs("Failed to enable UD0_COMPAT mode\n", ZYAN_STDERR);
            abort();
        }
    }
```

athre0z (approved):

PR #479, opened 2024-01-26
Enforce predicted instruction size (Fixes #478)

### LINKED ISSUE #478: Encoder fails to enforce predicted instruction size in some scenarios
`ZydisEncoderEncodeInstructionAbsolute` correctly predicts instruction sizes but if no relevant hint is provided it does nothing to enforce predicted instruction size. Usually this isn't an issue as size of immediate is enough to enforce correct variant. This however isn't sufficient for some edge cases where change of instruction size causes relative operand to fit into a smaller variant.

**Example:**
`jz` instruction is meant to be placed at `0x4000` and is supposed to jump to `0x4082`. 2-byte instruction (opcode + 8 bits of relative operand) can reach up to `0x4000+2+0x7F=0x4081`, so encoder predicts 6 bytes for jump instruction. Relative operand gets adjusted to `0x4082-(0x4000+6)=0x7C` but `0x7C <= 0x7F`, so encoder will encode 2-byte instruction without a proper hint, therefore resulting instruction doesn't jump where it should.

mappzor:
Fixes #478

### DIFF
--- a/src/Encoder.c
+++ b/src/Encoder.c
@@ -4449,6 +4449,22 @@ ZYDIS_EXPORT ZyanStatus ZydisEncoderEncodeInstructionAbsolute(ZydisEncoderReques
                     }
                     op->imm.s = rel;
                     adjusted_rel = ZYAN_TRUE;
+                    if (rel_info->accepts_scaling_hints == ZYDIS_SIZE_HINT_NONE)
+                    {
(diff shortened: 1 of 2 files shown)


flobernd (approved):
LGTM from my side. Thank you 🙂

PR #481, opened 2024-01-31
Keep same order of files for amalgamation

evdenis:
According to the documentation glob returns files in arbitrary order. Let's sort paths to keep the same order of files for amalgamation.


athre0z (approved):
LGTM, thanks! :) CI failure is unrelated.

PR #482, opened 2024-02-04
Make: support generating clangd DB

athre0z:
This adds a convenience make target for generating a clangd `compile_commands.json` database.


flobernd (approved):
LGTM!

PR #489, opened 2024-03-01
Add size and offset to immediate and memory operand structs

NaC-L:
Access to size of displacement value from operand

### DIFF
--- a/examples/Disassemble.c
+++ b/examples/Disassemble.c
@@ -76,7 +76,6 @@ int main(void)
 
         offset += instruction.length;
         runtime_address += instruction.length;
-
     }
 
     return 0;
--- a/examples/Disassemble.c
+++ b/examples/Disassemble.c
@@ -76,6 +76,7 @@ int main(void)
 
         offset += instruction.length;
         runtime_address += instruction.length;
+
     }
 
     return 0;
--- a/include/Zydis/DecoderTypes.h
+++ b/include/Zydis/DecoderTypes.h
@@ -154,14 +154,14 @@ typedef struct ZydisDecodedOperandMem_
          */
         ZyanI64 value;
         /**
-         * The physical displacement size, in bits.
-         */
-        ZyanU8 size;
-        /**
          * The offset of the displacement data, relative to the beginning of the
          * instruction, in bytes.
          */
         ZyanU8 offset;
+        /**
+         * The physical displacement size, in bits.
+         */
+        ZyanU8 size;
     } disp;
 } ZydisDecodedOperandMem;
 
--- a/include/Zydis/DecoderTypes.h
+++ b/include/Zydis/DecoderTypes.h
@@ -149,16 +149,19 @@ typedef struct ZydisDecodedOperandMem_
      */
     struct ZydisDecodedOperandMemDisp_
     {
-
         /**
          * The displacement value
          */
         ZyanI64 value;
-        
         /**
-        * The size of the displacement value
-        */
+         * The physical displacement size, in bits.
+         */
         ZyanU8 size;
+        /**
+         * The offset of the displacement data, relative to the beginning of the
+         * instruction, in bytes.
+         */
+        ZyanU8 offset;
     } disp;
 } ZydisDecodedOperandMem;
 
@@ -193,6 +196,15 @@ typedef struct ZydisDecodedOperandImm_
         ZyanU64 u;
         ZyanI64 s;
     } value;
+    /**
+      * The offset of the immediate data, relative to the beginning of the
+      * instruction, in bytes.
+      */
+    ZyanU8 offset;
+    /**
+     * The physical immediate size, in bits.
+     */
+    ZyanU8 size;
 } ZydisDecodedOperandImm;
 
 /**
--- a/include/Zydis/DecoderTypes.h
+++ b/include/Zydis/DecoderTypes.h
@@ -157,6 +157,11 @@ typedef struct ZydisDecodedOperandMem_
          * The displacement value
          */
         ZyanI64 value;
+        
+        /**
+        * size of the displacement value
+        */
+        ZyanU16 size;
     } disp;
 } ZydisDecodedOperandMem;
 
--- a/include/Zydis/DecoderTypes.h
+++ b/include/Zydis/DecoderTypes.h
@@ -149,19 +149,16 @@ typedef struct ZydisDecodedOperandMem_
      */
     struct ZydisDecodedOperandMemDisp_
     {
-        /**
-         * Signals, if the displacement value is used.
-         */
-        ZyanBool has_displacement;
+
         /**
          * The displacement value
          */
         ZyanI64 value;
         
         /**
-        * size of the displacement value
+        * The size of the displacement value
         */
-        ZyanU16 size;
+        ZyanU8 size;
     } disp;
 } ZydisDecodedOperandMem;
 
--- a/include/Zydis/SharedTypes.h
+++ b/include/Zydis/SharedTypes.h
@@ -47,7 +47,7 @@ extern "C" {
 /* ---------------------------------------------------------------------------------------------- */
 
 #define ZYDIS_MAX_INSTRUCTION_LENGTH    15
-#define ZYDIS_MAX_DISP_IMM_SIZE         32
+#define ZYDIS_MAX_DISP_IMM_SIZE         64
 #define ZYDIS_MAX_OPERAND_COUNT         10 // TODO: Auto generate
 #define ZYDIS_MAX_OPERAND_COUNT_VISIBLE  5 // TODO: Auto generate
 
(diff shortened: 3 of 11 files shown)


athre0z:
This information is already available in the public interface via `instruction->raw.disp.size`. Putting it where you are suggesting would admittedly be a bit nicer for discovery, but at the cost of 8 byte (after padding) times max number of operands of stack alloc. Our structs are quite large already and I'd prefer not to add to that.

NaC-L:
Hey, thanks for the response.

What about something like this?

```cpp
 struct ZydisDecodedOperandMemDisp_
    {
        /**
         * The displacement value
         */
        ZyanI64 value;
        
        /**
        * The size of the displacement value
        */
        ZyanU8 size;
    } disp;
```
    
If the size is 0, then it means no displacement value and we get rid of the `ZyanBool has_displacement;` so we can use the same amount of bytes. I understand the purpose of ZydisDecodedOperand and raw struct. However with this new way, replacing has_displacement with size we can use the same size and hold more info. The bytes are used anyways, why not make more use of them?

flobernd:
I would be ok with merging this. What do you think @athre0z?

We have a bunch of interface breaking changes anyways.

include/Zydis/DecoderTypes.h
```diff
@@ -149,14 +149,16 @@ typedef struct ZydisDecodedOperandMem_
...
-         */
-        ZyanBool has_displacement;
+
         /**
          * The displacement value
          */
         ZyanI64 value;
+        
+        /**
+        * The size of the displacement value
```
mappzor, line 159: Misaligned comment. Also there are extra blank lines before/after structure fields (inconsistent with rest of the header).

mappzor (changes requested):
This change should be mentioned in `assets/porting-guide-v4-v5.md`.

flobernd:
When we are on it, we should as well change the comment slightly to make it clear that this field contains the physical displacement size.

examples/Disassemble.c
```diff
@@ -76,6 +76,7 @@ int main(void)
 
         offset += instruction.length;
         runtime_address += instruction.length;
+
```
flobernd, line 79: This was probably a mistake?

athre0z:
Hmm yeah, that seems like a decent idea.

If we move the size into the operand, we should also move the offset field. I don't know how you'd reliably use the size without also having the offset. We should further apply the same change for immediate operands for consistency. 

Looking at `pahole` output:

```c
struct ZydisDecodedOperandMemDisp_ {
        ZyanBool                   has_displacement;     /*     0     1 */

        /* XXX 7 bytes hole, try to pack */

        ZyanI64                    value;                /*     8     8 */

        /* size: 16, cachelines: 1, members: 2 */
        /* sum members: 9, holes: 1, sum holes: 7 */
        /* last cacheline: 16 bytes */
};
```

my previous claim that adding a new field would increase struct size was in fact inaccurate: there are 7 padding bytes waiting to be put to use. In that sense moving the offset over as well is actually free. We could also keep `has_displacement`, but honestly there isn't really any good reason: using `size` for both seems more elegant.

examples/Disassemble.c
NaC-L: whoops

include/Zydis/SharedTypes.h
```diff
@@ -47,6 +47,7 @@ extern "C" {
 /* ---------------------------------------------------------------------------------------------- */
 
 #define ZYDIS_MAX_INSTRUCTION_LENGTH    15
+#define ZYDIS_MAX_DISP_IMM_SIZE         32
```
flobernd, line 50: This should be `64` to support `mov{abs}` and `jmpabs` (as soon as APX is merged).

tools/ZydisFuzzShared.c
```diff
@@ -175,7 +175,7 @@ void ZydisValidateEnumRanges(const ZydisDecodedInstruction* insn,
             ZYDIS_CHECK_ENUM(op->mem.segment, ZYDIS_REGISTER_MAX_VALUE);
             ZYDIS_CHECK_ENUM(op->mem.base, ZYDIS_REGISTER_MAX_VALUE);
             ZYDIS_CHECK_ENUM(op->mem.index, ZYDIS_REGISTER_MAX_VALUE);
-            ZYDIS_CHECK_ENUM(op->mem.disp.has_displacement, ZYAN_TRUE);
+            ZYDIS_CHECK_ENUM(op->mem.disp.size, ZYDIS_MAX_DISP_IMM_SIZE);
```
flobernd, line 178: Not sure if `ZYDIS_CHECK_ENUM()` is appropriate here, as the size has definition holes.

@mappzor Should we as well add checks for `disp.offset`, `imm.offset` and `imm.size`?

flobernd (commented):
Thanks for updating the PR 🙂 I left some more comments.

tools/ZydisFuzzShared.c
mappzor: I would go as strictly as possible:

- displacement size must be one of 0, 8, 32
- immediate size must be one of 0, 8, 16, 32, 64
- offset + (size / 8) <= instruction length

tools/ZydisFuzzShared.c
```diff
@@ -147,6 +147,67 @@ void ZydisValidateEnumRanges(const ZydisDecodedInstruction* insn,
             " = 0x%016" PRIX64 "\n", (ZyanU64)(value), (ZyanU64)(max));                            \
         abort();                                                                                   \
     }
+#   define ZYDIS_CHECK_MAX ZYDIS_CHECK_ENUM
```
NaC-L, line 150: I'm not a fan of these macros but this was the only solution i could come up with.

tools/ZydisFuzzShared.c
```diff
@@ -147,6 +147,67 @@ void ZydisValidateEnumRanges(const ZydisDecodedInstruction* insn,
...
+        abort();                                                                                   \
+    }
+
+
+
+#define ZYDIS_CHECK_VALUE_ONE_OF_N(value, ...) \
+        GET_MACRO(__VA_ARGS__, ZYDIS_CHECK_VALUE_ONE_OF_5, ZYDIS_CHECK_VALUE_ONE_OF_4, ZYDIS_CHECK_VALUE_ONE_OF_3, ZYDIS_CHECK_VALUE_ONE_OF_2, ZYDIS_CHECK_VALUE_ONE_OF_1)(value, __VA_ARGS__)
+
+#define GET_MACRO(_1, _2, _3, _4, _5, NAME, ...) NAME
+
```
NaC-L, line 210: I'm not a fan of these macros but this was the only solution i could come up with.

tools/ZydisFuzzShared.c
flobernd: @mappzor Because of technical implementation details, we currently have to accept `0, 8, 16, 32, 64` for DISP as well, because I've choosen to use DISP to map the MOFFS operand type:

```
A1 12 34 56 78 12 34 56 78
:  :..DISP
:..OPCODE
```

tools/ZydisFuzzShared.c
flobernd: Thanks for taking the effort to create a generic implementation. For me personally, I would have been fine with just a manual check without a macro as well. I let @mappzor decide about this, as it's mostly the codebase he worked on.

mappzor: I would get rid of macros for this one. A small utility function that checks against known goods (0, 8, 16, 32, 64) would do the trick. It doesn't even have to accept array of good values as parameter, they could be hardcoded as static array within the function. If we ever need more flexiblity we'll just update the function when it's actually needed.

tools/ZydisFuzzShared.c
```diff
@@ -136,6 +136,14 @@ void ZydisPrintInstruction(const ZydisDecodedInstruction* instruction,
 
 #endif
 
+void ZydisValidateImmediateSize(ZyanU64 value) {
+    if (value != 0 && value != 8 && value != 16 && value != 32 && value != 64) {
```
flobernd, line 140:
```suggestion
void ZydisValidateImmediateSize(ZyanU64 value)
{
    if ((value != 0) && (value != 8) && (value != 16) && (value != 32) && (value != 64))
    {
```

Looks good now! Just another small suggestion regarding code style to be consistent with the rest of the code.

flobernd:
LGTM from my side! Thanks 🙂

flobernd (approved):

flobernd:
@athre0z @mappzor Are you fine with merging this as well?

athre0z (approved):
Yes, I think this is a good change. For me personally those two fields are the only thing in the entire `raw` portion that I ever use. Going forward we could consider moving `raw` into the decoder context instead, or make outputting it optional: I suspect that barely anything aside from `ZydisInfo` uses the other stuff. Either way, just thinking out loud here: that's a problem for another day. Thanks for the effort @NaC-L, LGTM!

mappzor:
LGTM!

PR #499, opened 2024-04-03
remove unnecessary check EVEX.V' field

OptionalAssistant:
We have  already checked this condition above no need to check again


flobernd (approved):
Good catch!

flobernd:
@OptionalAssistant I think you accidentally pushed an optional newline a few hours ago.

PR #500, opened 2024-04-06
Fix issue with `llvm-rc` when `ZYDIS_BUILD_SHARED_LIB` is enabled

netadr:
In addition to the changes mentioned in zyantific/zycore-c#70, this changes the encoding of `resources/VersionInfo.rc` from UTF-16LE to UTF-8 as `clang`'s preprocessor does not support UTF-16.


flobernd (commented):
Thanks for the PR!

Shouldn't we replace the copyright sign here as well?

I merged your change in zycore-c. Maybe you could update the submodule to the latest version as well in this PR.

netadr:
> Shouldn't we replace the copyright sign here as well?

I changed it already here as well.

Sorry, GitHub doesn't do a great job of showing this due to the encoding change, maybe I should have clarified!

flobernd:
@netadr Oh, my bad. I even clicked on "view file", but GitHub seems to show me the old version.

Do you want to update the zycore-c submodule in this PR? We can do this in a later step as well, if that's more comfortable for you.

netadr:
> Do you want to update the zycore-c submodule in this PR? We can do this in a later step as well, if that's more comfortable for you.

@flobernd *Should* be fixed now!

flobernd (approved):
Looks good! Thanks again.

athre0z (approved):
👍

PR #507, opened 2024-05-31
Fix zydis not compiling under -Werror

supernova9sky:
Added an explicit cast in the assert to silence the "-Wsign-compare" error.

### DIFF
--- a/src/Register.c
+++ b/src/Register.c
@@ -175,7 +175,7 @@ ZydisRegister ZydisRegisterGetLargestEnclosing(ZydisMachineMode mode, ZydisRegis
         [ZYDIS_REGCLASS_FLAGS] = { ZYDIS_REGISTER_FLAGS, ZYDIS_REGISTER_EFLAGS, ZYDIS_REGISTER_RFLAGS },
         [ZYDIS_REGCLASS_IP   ] = { ZYDIS_REGISTER_IP   , ZYDIS_REGISTER_EIP   , ZYDIS_REGISTER_RIP    },
     };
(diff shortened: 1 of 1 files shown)


flobernd (approved):
LGTM! Thank you!

src/Register.c
```diff
@@ -175,7 +175,7 @@ ZydisRegister ZydisRegisterGetLargestEnclosing(ZydisMachineMode mode, ZydisRegis
         [ZYDIS_REGCLASS_FLAGS] = { ZYDIS_REGISTER_FLAGS, ZYDIS_REGISTER_EFLAGS, ZYDIS_REGISTER_RFLAGS },
         [ZYDIS_REGCLASS_IP   ] = { ZYDIS_REGISTER_IP   , ZYDIS_REGISTER_EIP   , ZYDIS_REGISTER_RIP    },
     };
-    ZYAN_ASSERT(reg_class < ZYAN_ARRAY_LENGTH(STATIC_MAPPING));
+    ZYAN_ASSERT((ZyanU64)reg_class < ZYAN_ARRAY_LENGTH(STATIC_MAPPING));
```
athre0z, line 178: Shouldn't this be a cast to `ZyanUSize` instead? Otherwise I imagine that it'd still generate a warning in 32 bit builds.

supernova9sky: Thanks. 
It probably should be, because ZyanUSize (which ZYAN_ARRAY_LENGHT returns) and ZyanU64 can differ in size.

athre0z: Ok, cool -- can you change the PR accordingly, please?

supernova9sky: Yes. Changed it. Thank you for your time :)

athre0z:
Thanks, LGTM now! :)

PR #508, opened 2024-06-01
Use alias targets

dg0yt:
Complements and depends on https://github.com/zyantific/zycore-c/pull/71.
Fixes build with "system" zycore.


flobernd:
Let's update the submodule in this PR as well as soon as the related patch is merged.

flobernd (approved):
LGTM! Thanks again!

PR #516, opened 2024-07-10
CMake: don't unconditionally compile Segment.{h,c}

athre0z:
As pointed out by @korli [here](https://github.com/zyantific/zydis/commit/dbaec9104745e1869926358a26586d9ccf657666#r143977805), we previously accidentally compiled Segment.h / Segment.c regardless of whether `ZYDIS_FEATURE_SEGMENT` was enabled or not.


flobernd (approved):
LGTM!

PR #520, opened 2024-08-06
Utils: Handle OPERAND_TYPE_REGISTER in ZydisCalcAbsoluteAddressEx.

waddlesplash:
This way, consumers of the API can call the method with rather than having to special-case `ZYDIS_OPERAND_TYPE_REGISTER`. Found & implemented for Haiku's Debugger (which now uses Zydis) to perform stepping operations properly.


flobernd (approved):
LGTM! Thank you 🙂

PR #533, opened 2024-11-05
Encoder: APX support

mappzor:


flobernd (approved):
Great work as usual 🙂 Thanks a lot!

PR #546, opened 2025-01-07
Bump version number to 5.0.0

mochaaP:
Since we already introduced breaking API changes, so there should be a way to detect those changes for users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.15")
     cmake_policy(SET CMP0091 NEW)
 endif ()
 
-project(Zydis VERSION 4.0.0.0 LANGUAGES C)
+project(Zydis VERSION 5.0.0.0 LANGUAGES C)
 
 include(GNUInstallDirs)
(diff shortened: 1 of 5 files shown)


mochaaP:
Unrelated but worth mentioning:

I wish we could have a tagged release for v4 with meson, so I could add Zydis to meson's WrapDB without waiting for v5 release :)

flobernd:
@mappzor and me worked on the APX branch a while ago and our work is mostly finished by now. 

We have to polish some of the APIs before we can trigger the V5 release.

I hope to find some time for this soon. Not sure how much effort it would be to backport the Meson support to V4.

mochaaP:
I'll make a new PR for the v4 meson support.

flobernd:
Thank you! I'll then make sure to create a v4 maintenance branch first, in which we should merge your PR and backport any other non breaking fix from master until V5 gets released.

athre0z:
I think we should still bump the version already, even if we don't release v5 yet. I think that's how most projects do it: once you introduce the first breaking change and fork off a maintenance branch for current major version, you bump version on `main` to let everyone know about the breaking changes.

flobernd (approved):

Issue #293, opened 2021-12-17
Encoder requires xchg to have memory operand first


ZehMatt:
The decoder table has no form of ```reg, [mem]``` so naturally it fails to find a definition.

PR #561, opened 2025-03-19
Fixed `xchg` aliasing (Fixes #560)

### LINKED ISSUE #560: Inconsistent Re-encoding of `xchg eax, eax` to `nop` in 64-bit Mode (`ZYDIS_MACHINE_MODE_LONG_64`)
I am currently working on decoding the instruction `xchg eax, eax` (encoding: `87 C0`) in `ZYDIS_MACHINE_MODE_LONG_64` mode and subsequently re-encoding it. However, the re-encoding process results in the instruction `nop` (encoding: `90`), which is inconsistent with the expected behavior.

During testing using the ZydisFuzzReEncoding tool, the following output was observed:
```
-64 -64 87C0 xchg eax, eax
-64 -64 90 nop
```

Upon further investigation, I noticed that this specific case is already included in the test suite, as shown in the test data here:  
https://github.com/zyantific/zydis/blob/bffbb610cfea643b98e87658b9058382f7522807/tests/re_enc_test_cases.json#L224-L229

However, the issue appears to be intentionally ignored in the validation logic, as evidenced by the following code segment:  
https://github.com/zyantific/zydis/blob/bffbb610cfea643b98e87658b9058382f7522807/tools/ZydisFuzzShared.c#L247-L264

This behavior is problematic in a 64-bit architecture. For example, if we assume `rax=0xFFFFFFFFFFFFFFFF`, executing `xchg eax, eax` results in `rax=0x00000000FFFFFFFF`. This occurs because writing to the `eax` register clears the upper 32 bits of the `rax` register. In contrast, executing `nop` leaves `rax` unchanged at `0xFFFFFFFFFFFFFFFF`. As a result, the two instructions produce inconsistent outcomes, which is not desirable.

**Proposed Solution:**  
To address this issue, I suggest adding constraints to the encoding rules to ensure that the simplification of `xchg eax, eax` to `nop` is only applied in 32-bit architectures. By restricting this optimization to 32-bit mode, the issue would be avoided in 64-bit mode, preserving the expected behavior of the `xchg eax, eax` instruction.

mappzor:
Fixes #560

Currently using `swappable` mechanism is the best possible solution. Once we have new generator a data-only solution should be possible. It's very similar issue as with #293. Generating dangling decoder nodes just so encoder could reference them would easily solve both issues.

Relevant changes in behavior:

| Requested | Old | New |
| --- | --- | --- |
| M16:eax:eax | 66 90 nop | 90 nop |
| M32:ax:ax | 66 90 nop | 90 nop |
| M64:ax:ax | 66 90 nop | 90 nop |
| M64:eax:eax | 90 nop | 87 C0 xchg eax, eax |
| M64:rax:rax | 48 90 nop | 90 nop |


flobernd (approved):
LGTM! Thank you 🙂

PR #573, opened 2025-05-26
fixing unaligned reads of the size table

jakubDoka:
This esentially replaces the direct read from the size table to unaligned read (align 1) because most likely the struct containing the table has an odd byte size.


jakubDoka:
I checked the disasm and GCC at least can optimize this to:
```asm
movzx   eax, WORD PTR [rdi+rsi*2]
```
Can't really get better then this

mappzor:
Superseded by #582

PR #577, opened 2025-06-09
Remove typecasts from version macros 

ksco:
Using typecasts in version macros prevents them from being used in the #if directive. According to the GCC documentation [1], the #if directive only allows integer constant expressions, and typecasts are not permitted in such expressions.

This change removes the typecasts from the version macros, allowing them to be used in preprocessor conditionals.

[1] https://gcc.gnu.org/onlinedocs/cpp/If.html


flobernd:
Closed in favor of #546.

Thank you for the PR anyways 🙂

PR #595, opened 2025-12-09
build/meson: remove workaround, revise exports

mochaaP:
- **build/meson: remove msvc b_lto workaround**
  tracked in https://github.com/mesonbuild/meson/pull/15357

- **build/meson: revise exports**
  split export static and shared meson targets under the same name
  (no longer needed to manually `-DZYDIS_STATIC_BUILD` in downstream projects, needs https://github.com/zyantific/zycore-c/pull/95)
  make pkgconfig optional
  export predef cflags

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -33,10 +33,7 @@ nolibc = get_option('nolibc')
 
 # Auto feature set
 encoder = encoder.enable_auto_if(
-  minimal.disabled()
-  and decoder.allowed()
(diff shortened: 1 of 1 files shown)


mochaaP:
The revised export scheme currently has a bug in upstream meson: https://github.com/mesonbuild/meson/issues/15371
a current workaround is to ensure default_library and prefer_static is the same

PR #598, opened 2026-02-26
Fix ZydisPerfTest on NetBSD.

alnsn:
Do a bare minimum to fix the build. Don't try to use cpuset API and pthread_setaffinity_np() because they require super-user privileges (or setting security.models.extensions.user_set_cpu_affinity).

Relevant link: https://man.netbsd.org/NetBSD-10.1/affinity.3


flobernd (approved):
LGTM, thank you!

Issue #600, opened 2026-03-30
CMake warnings during configuration


kerrytazi:
Minimal example:
```cmake
cmake_minimum_required(VERSION 3.10)
project("zydis_test")

include(FetchContent)

FetchContent_Declare(
	Zydis
	GIT_REPOSITORY https://github.com/zyantific/zydis.git
	GIT_TAG        v4.1.1
	GIT_SHALLOW    TRUE
)
FetchContent_MakeAvailable(Zydis)

get_target_property(zydis_source_dir Zydis SOURCE_DIR)
add_executable(zydis_test ${zydis_source_dir}/examples/DisassembleSimple.c)
target_link_libraries(zydis_test PRIVATE Zydis)
```

CMake output:
```
CMake Deprecation Warning at build/_deps/zydis-src/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


CMake Deprecation Warning at build/_deps/zydis-src/dependencies/zycore/CMakeLists.txt:5 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```

I'd like not to see any warnings

`cmake version 4.3.0`

mappzor:
You are pulling version 4.1.1 of Zydis which required CMake 3.9. We currently require 3.10 and that's enough to get rid of that warning. This issue can be considered as resolved via #569.

PR #601, opened 2026-03-30
Fixed cmake warnings. Related to zyantific/zydis#600

kerrytazi:


mappzor:
Change isn't necessary here (see https://github.com/zyantific/zydis/issues/600#issuecomment-4152509092). Also every CMake version bump needs to start with Zycore before Zydis's version and submodule can be updated.

Issue #263, opened 2021-11-06
Consider reworking `ZydisShortString`


athre0z:
Because #259 had to remove the struct packing on aarch64 darwin, it might be worth considering to do what @flobernd suggested in the PR:

> Maybe as an improvement, we could prepend the length to the actual strings (this should be easily possible as we generate the strings anyways) and remove the 'length' field from the struct. In this case a struct would not even be necessary, but probably still desired as a hint.

thestr4ng3r:
I am currently taking a look at implementing this for support on OpenBSD/sparc64, where the `#pragma pack(push, 1)` also causes issues.
These would be two options:
```c
typedef struct ZydisShortString_
{
    ZyanU8 size;
    const char data[];
} ZydisShortString;

static const ZydisShortString STR_REGISTERS_VALUE_0 = ZYDIS_MAKE_SHORTSTRING("none");

static const ZydisShortString * STR_REGISTERS[] =
{
    &STR_REGISTERS_VALUE_0,
    // ...
}
```

Or shorter:

```c
static const char * STR_REGISTERS[] =
{
    "\x04none",
    // ...
}
```

Which would you prefer? Or do you have something else in mind?

Currently I am struggling to figure out how `src/Generated/EnumMnemonic.inc` and `src/Generated/FormatterStrings.inc` are generated as there is no reference in https://github.com/zyantific/zydis-db to those at the moment. `src/Generated/EnumRegister.inc` in contrast can be easily reproduced and adjusted with the https://github.com/zyantific/zydis-db/blob/df7bbef6239cf74c0bf33043f46ef2c0bc32b26c/scripts/gen_registers.py script.
Were the other files so far maybe only produced by the legacy Pascal one? Should the Python script be extended to also generate those or should it all be done in the C# project?

flobernd:
Hi @thestr4ng3r ,

thanks for working on this!

The FormatterStrings.inc generation indeed is not yet ported over from the legacy Delphi project to the C# one.

The goal is to have everything in the C# project at some point 🙂

PR #603, opened 2026-05-06
Fix ZydisEmitUInt with size 1 on big endian

thestr4ng3r:
Even if only a single byte is copied, that byte will not be at the lowest address of a ZyanU64 value, so a cast to ZyanU8 is still necessary, only without a byte swap.

Tested on OpenBSD/sparc64. More changes will be necessary to get zydis to work there properly, in particular the packing of `ZydisShortString` causes segfaults. But the approach suggested in #263 is probably the best way to fix it for sparc since it has very strict alignment requirements.

### DIFF
--- a/src/Encoder.c
+++ b/src/Encoder.c
@@ -3382,11 +3382,11 @@ static ZyanStatus ZydisEmitUInt(ZyanU64 data, ZyanU8 size, ZydisEncoderBuffer *b
 
     // The size variable is not passed on purpose to allow the compiler
     // to generate better code with a known size at compile time.
+#if ZYAN_ENDIAN == ZYAN_LITTLE_ENDIAN
     if (size == 1)
     {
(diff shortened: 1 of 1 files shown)


src/Encoder.c
```diff
@@ -3400,6 +3400,11 @@ static ZyanStatus ZydisEmitUInt(ZyanU64 data, ZyanU8 size, ZydisEncoderBuffer *b
         ZYAN_MEMCPY(buffer->buffer + buffer->offset, &data, 8);
     }
 #else
+    if (size == 1)
+    {
+        ZyanU8 value = data;
```
mappzor: Please add explicit cast to `ZyanU8` here.

mappzor (changes requested):

mappzor (approved):

mappzor:
LGTM, thank you!

PR #604, opened 2026-05-13
Rework ZydisShortString (Fixes #263)

### LINKED ISSUE #263: Consider reworking `ZydisShortString`

thestr4ng3r:
This corresponds to https://github.com/zyantific/zydis-db/pull/43

The new structure includes the length directly before the string in memory, similar to Pascal strings. The advantage is that this is naturally packed without explicit compiler directives. The explicit packing pragma previously caused problems at least on recent macOS and OpenBSD/sparc64.

My first attempt was something like this:
```c
typedef const void ZydisShortString;
ZYAN_INLINE ZyanU8 ZydisShortStringGetLength(ZydisShortString* str)
{
       return ((const ZyanU8*)str)[0];
}

ZYAN_INLINE const char *ZydisShortStringGetString(ZydisShortString* str)
{
       return ((const char*)str) + 1;
}
```
and then generating all shortstring literals like `"\x07example"`. This would have worked but also required a bit more public header changes and to also to port generation of `src/Generated/FormatterStrings.inc` to C# first in addition to enum generation.
Since both approaches end up with the same memory layout, I went with the struct in the end.

A small change in the generator causes that some trailing spaces have been removed, which causes diffs also in generated files other than the `Enum*` ones.

Fixes #263


flobernd (approved):
LGTM, thank you 🙂

flobernd:
Just noticed compilation fails with:

```
error: flexible array initialization is a GNU extension [-Werror,-Wgnu-flexible-array-initializer]
```

Any way we can redesign this to be fully standard compliant? @thestr4ng3r

thestr4ng3r:
That is indeed a problem, I have to think about this a bit more. But likely I will have to first port `src/Generated/FormatterStrings.inc` generation to C# since it might not be possible to do what we want without altering the `static const ZydisShortString STR_ADD = ZYDIS_MAKE_SHORTSTRING("+");` lines in any way.

In case the zydis-db repo should stay fully in sync with zydis until then, the single commit https://github.com/zyantific/zydis-db/commit/f714efb293995a800112b9f9ad63241237a65478 could be reverted.

flobernd:
@thestr4ng3r I approved and merged your follow up in the generator repo. Thank you! We can keep the previous commit for now.

thestr4ng3r:
@flobernd I went with a similar approach that is used for the `ZydisPredefinedToken` now where the struct contains a flexible array member, but initialization is done by using an inline struct with a fixed size array and then casting to `ZydisShortString*`:
```c
/// ZYDIS_DEFINE_SHORTSTRING_DATA(HELLO, 5, "hello"); expands to
static const struct ZydisShortString_HELLO_
{
    ZyanU8 size;
    const char data[5 + 1];
} STR_DATA_HELLO = { 5, "hello" };
```

```c
// ZYDIS_SHORTSTRING(HELLO) expands to
((const ZydisShortString*)&STR_DATA_HELLO)
```
corresponding generator changes: https://github.com/zyantific/zydis-db/pull/47

include/Zydis/ShortString.h
```diff
@@ -58,31 +60,46 @@ extern "C" {
...
+ * Defines a constant variable of an inline struct that is compatible to `ZydisShortString`.
  *
+ * @param   name    Name for the symbols, prefixed by STR_DATA_, will also be used as a suffix for
+ *                  the inline struct name.
+ * @param   len     Length of string in bytes, excluding the trailing 0 terminator.
  * @param   string  The C-string constant.
  */
-#define ZYDIS_MAKE_SHORTSTRING(string) \
-    { string, sizeof(string) - 1 }
+#define ZYDIS_DEFINE_SHORTSTRING_DATA(name, len, string) \
```
mappzor: I'd prefer to keep old name for the macro and change its definition a bit:
```c
#define ZYDIS_MAKE_SHORTSTRING(name, string) \
    static const struct ZydisShortString_ ## name ## _ \
    { \
        ZyanU8 size; \
        const char data[sizeof(string)]; \
    } STR_DATA_ ## name = { sizeof(string) - 1, string } \
```
This eliminates the need for explicit string size.

mappzor (changes requested):
Thank you for working on this. After addressing small change to the macro and resolving merge conflicts this should be ready to go.

include/Zydis/ShortString.h
thestr4ng3r: Ok, adjusted the macro.

thestr4ng3r:
Tests currently fail because https://github.com/zyantific/zydis-db/commit/e5d2961cc075861c8d99c9f88c9458942b2d3140 is already in zydis-db master, and thus the generated code in this pr is based on it, but zydis master is not generated with it.

mappzor:
That was to be expected. Please just rebase the tests and commit the result, in `tests` directory there's a script and you run it with: `py regression.py rebase <path to ZydisInfo binary>`.

mappzor (approved):
LGTM, thank you!

PR #611, opened 2026-05-23
Decoder: remove dead IP_ASZ implicit-register type

dxbjavid:
No operand in the database uses `ZYDIS_IMPLREG_TYPE_IP_ASZ`, so its branch in `ZydisDecodeOperandImplicitRegister` is unreachable. Removes the enum value and the dead switch case. `IP_SSZ` keeps its existing EIP mapping.


mappzor:
CALL documentation:
> For a near call absolute, an absolute offset is specified indirectly in a general-purpose register or a memory location
(r/m16, r/m32, or r/m64). The operand-size attribute determines the size of the target operand (16, 32 or 64
bits). When in 64-bit mode, the operand size for near call (and all near branches) is forced to 64-bits. Absolute
offsets are loaded directly into the EIP(RIP) register. **If the operand size attribute is 16, the upper two bytes of the
EIP register are cleared, resulting in a maximum instruction pointer size of 16 bits.**

Documentation specifically makes a point about EIP/RIP being used at all times but never IP. XED outputs also reflect that philosophy, 16 and 32-bit outputs refer to EIP and 64-bit to RIP.

I can see the argument for change but I'm not sure if it's worth. A lot of hidden registers for complex operations such as `iret`, `vmcall`, etc. provide minimum information about semantics as their full functionality cannot be expressed in a simplified way. Same applies here, on one hand it's an IP modification but on the other high EIP bytes get modified as well. Changing this would be a divergence from XED and it's a breaking change for code which relied on `reg == ZYDIS_REGISTER_EIP || reg == ZYDIS_REGISTER_RIP`. I think I'm leaning towards keeping it as is.

> asymmetric with the neighbouring IP_ASZ and FLAGS_SSZ tables in the same function

`case ZYDIS_IMPLREG_TYPE_IP_ASZ` is dead code as no `asz_ip` operands exist in the database right now. Perhaps I'd better to remove that case instead of changing `case ZYDIS_IMPLREG_TYPE_IP_SSZ`.

@flobernd Any thoughts on this?

flobernd:
@mappzor Yes, we kept EIP on purpose here - let's not change it 🙂 

Removing ZYDIS_IMPLREG_TYPE_IP_ASZ also sounds good! I think that's a leftover from a very early version that incorrectly used ASZ for IP registers.

dxbjavid:
Understood, thanks for the clarification from both of you.
Keeping  EIP for  IP_SSZ makes sense given the architectural semantics and XED compatibility. I agree that removing the dead  'IP_ASZ' path is probably the cleaner follow-up.

dxbjavid:
Repurposed this to just drop the dead IP_ASZ type and its handler, and left IP_SSZ on EIP as you decided. Builds clean and the regression tests still pass since nothing referenced it. If you'd rather fold it into your own cleanup, feel free to close this out.

mappzor (approved):
LGTM

TODO: We should remove it from generator project as well.

PR #614, opened 2026-05-26
Decoder: fix uses_egpr for R20/R28 used as SIB index

dxbjavid:
ZydisPopulateRegisterIds derives has_index from the raw 3-bit sib.index field, so an extended GPR index whose low three bits are 100b (R20 and R28) is treated as the no-index form and apx.uses_egpr is left clear, even though the operand decodes to an EGPR. The base path already keys off the full id_base, so the index test should use id_index too. After the change ZydisInfo on D5 28 03 04 20 reports USES_EGPR: Y for the r20 index.


mappzor (approved):
LGTM, this issue is a regression introduced via #582

flobernd (approved):
LGTM! Yes, that's an APX oversight 🙂

PR #615, opened 2026-05-27
truncate 32-bit relative branch target in ZydisCalcAbsoluteAddress

dxbjavid:
Found while fuzzing ZydisCalcAbsoluteAddress with high runtime
addresses. For a 32-bit-operand near branch in 16/32-bit modes the
result is never truncated to the IP width, so a rel32 at the top of
the 4 GiB space (rip 0xFFFFFFF0) returns 0x100000015 instead of the
wrapped EIP 0x15. The 16-bit arm already masks to 0xFFFF; do the
same for 32-bit.

### DIFF
--- a/src/Utils.c
+++ b/src/Utils.c
@@ -110,6 +110,10 @@ ZyanStatus ZydisCalcAbsoluteAddress(const ZydisDecodedInstruction* instruction,
                 {
                     *result_address &= 0xFFFF;
                 }
+                else if (instruction->operand_width == 32)
+                {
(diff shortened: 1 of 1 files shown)


mappzor:
> result is never truncated to the IP width

It's responsibility of the caller to use as many bits as required for desired address width (e.g. our formatters do this). Truncation is guaranteed only for memory operands with absolute addresses and this behavior is appropriately documented.

Changing this behavior would require more changes in the codebase and it'd be a change of existing API contract. It doesn't seem necessary at the moment.

PR #616, opened 2026-05-27
Encoder: bound operand count in DecodedInstructionToEncoderRequest

dxbjavid:
ZydisEncoderDecodedInstructionToEncoderRequest writes operand_count_visible operands into the fixed request->operands[ZYDIS_ENCODER_MAX_OPERANDS] array but only validates the count against instruction->operand_count_visible, never the array bound. A caller-supplied ZydisDecodedInstruction with operand_count_visible > 5 then overruns the array and writes past the request struct. Reject it up front, matching the cap ZydisEncoderCheckRequestSanity already enforces on request->operand_count.

### DIFF
--- a/src/Encoder.c
+++ b/src/Encoder.c
@@ -4883,7 +4883,8 @@ ZYDIS_EXPORT ZyanStatus ZydisEncoderDecodedInstructionToEncoderRequest(
         ZyanU8 operand_count_visible, ZydisEncoderRequest *request)
 {
     if (!instruction || !request || (operand_count_visible && !operands) ||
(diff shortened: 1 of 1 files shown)


mappzor:
`ZydisDecodedInstruction` is considered as trusted "immutable object" in this context (always comes from previous calls to decoder, it should be treated as read-only after decoding and never modified manually) making existing check `operand_count_visible != instruction->operand_count_visible` sufficient. This lightweight conversion function does minimal run-time sanity checks to prevent common user errors when passing arguments, malicious modification of decoded instruction doesn't fall into this category hence it's out of scope.

PR #617, opened 2026-05-27
Formatter: keep disp_padding preset when set to PADDING_AUTO

dxbjavid:
A DISP_PADDING of ZYDIS_PADDING_AUTO is resolved to the style preset in ZydisFormatterSetProperty, then immediately overwritten by the unconditional disp_padding = value, so it stays at -1. The AT&T/Intel displacement printers pass disp_padding straight to the number formatter with no AUTO handling (unlike imm/addr), so SetProperty returns SUCCESS but every later format of a memory operand with a displacement fails with INSUFFICIENT_BUFFER_SIZE. Move the assignment into an else.

### DIFF
--- a/src/Formatter.c
+++ b/src/Formatter.c
@@ -239,7 +239,8 @@ ZyanStatus ZydisFormatterSetProperty(ZydisFormatter* formatter, ZydisFormatterPr
     }
     case ZYDIS_FORMATTER_PROP_DISP_PADDING:
     {
-        if (value > 0xFF)
+        // Unlike imm/addr padding, displacement padding does not support `ZYDIS_PADDING_AUTO`
+        if (((ZydisPadding)value == ZYDIS_PADDING_AUTO) || (value > 0xFF))
         {
             return ZYAN_STATUS_INVALID_ARGUMENT;
         }
--- a/src/Formatter.c
+++ b/src/Formatter.c
@@ -239,15 +239,7 @@ ZyanStatus ZydisFormatterSetProperty(ZydisFormatter* formatter, ZydisFormatterPr
     }
     case ZYDIS_FORMATTER_PROP_DISP_PADDING:
     {
-        if ((ZydisPadding)value == ZYDIS_PADDING_AUTO)
-        {
-            if ((ZyanUSize)formatter->style > ZYDIS_FORMATTER_STYLE_MAX_VALUE)
-            {
-                return ZYAN_STATUS_INVALID_ARGUMENT;
-            }
-            formatter->disp_padding = FORMATTER_PRESETS[formatter->style]->disp_padding;
-        }
-        else if (value > 0xFF)
+        if (value > 0xFF)
         {
             return ZYAN_STATUS_INVALID_ARGUMENT;
         }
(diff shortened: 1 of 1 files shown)


mappzor:
It should be noted that `ZYDIS_FORMATTER_PROP_IMM_PADDING` is handled in the same manner. Putting value overwrite problem aside for a moment let's look at the documentation:

```
    /**
     * Controls the padding of displacement values.
     *
     * Pass `ZYDIS_PADDING_DISABLED` to disable padding, or any other integer value for custom
     * padding.
     */
    ZYDIS_FORMATTER_PROP_DISP_PADDING,
    /**
     * Controls the padding of immediate values.
     *
     * Pass `ZYDIS_PADDING_DISABLED` to disable padding, `ZYDIS_PADDING_AUTO` to pad all
     * immediates to the operand-width (hexadecimal only), or any other integer value for custom
     * padding.
     */
    ZYDIS_FORMATTER_PROP_IMM_PADDING,
```

For `ZYDIS_FORMATTER_PROP_DISP_PADDING` there's no mention of `ZYDIS_PADDING_AUTO` at all, so `ZYDIS_PADDING_AUTO` processing inside `ZydisFormatterSetProperty` should not exist.
For `ZYDIS_FORMATTER_PROP_IMM_PADDING` passing `ZYDIS_PADDING_AUTO` will first use value taken from formatter style preset and overwrite it with `ZYDIS_PADDING_AUTO`. Actually this overwrite saves the day because `ZYDIS_PADDING_AUTO` value will eventually reach `ZydisFormatterBasePrintIMM` which handles `ZYDIS_PADDING_AUTO` as intended.

Unless I'm misunderstanding the intent here, both cases of "resolve using style preset" are unintended leftovers and should be removed.

dxbjavid:
You're right, the preset resolution was a leftover in both. Reworked it: DISP_PADDING no longer special-cases AUTO at all, so passing it now gets rejected like any other out-of-range value (matches the docs, and the disp printer doesn't understand AUTO anyway). IMM_PADDING keeps AUTO but just passes it straight through to the field the same way ADDR_PADDING_* already do, since ZydisFormatterBasePrintIMM resolves it. Regression suite still passes.

dxbjavid:
Hi @mappzor 

Thank you again for the review and feedback. I have addressed the points raised and pushed the updated implementation.

When you have a chance, could you please take another look and let me know if anything else needs to be adjusted?

Thanks!

flobernd:
> Unless I'm misunderstanding the intent here, both cases of "resolve using style preset" are unintended leftovers and should be removed.

I think you are right. The preset is applied in `ZydisFormatterInit` and due to the override the "auto" block here is a no-op anyways. Removing it is the right call.

Oh WAIT:
I _think_ for DISP we should actually keep the block, but fix the logic. Since DISP does not support AUTO, we should use the value from the presets instead (or probably better: INVALID_ARGUMENT if the user tries to pass AUTO here). The PADDING enum is used in multiple places, but only DISP does not support AUTO.

flobernd (approved):

dxbjavid:
Agreed, INVALID_ARGUMENT is the right behaviour for disp. As it happened AUTO was already getting rejected, since value is a ZyanUPointer the -1 wraps above 0xFF and tripped the existing range check, but that was a bit too implicit to rely on. I've added an explicit AUTO check on the disp case so the intent is clear, with a short comment noting disp doesn't support AUTO. Regression suite still passes.

mappzor (approved):

dxbjavid:
TY

PR #623, opened 2026-05-29
strip NF bit from EVEX aaa mask selector for APX NF instructions

dxbjavid:
Decoding an APX `NF` instruction (e.g. `62 F4 FC 0C 01 C0`, NF form of `add`) in `ZydisSetAVXInformation` reads `raw.evex.aaa` as the opmask selector, but for `NF` instructions bit 2 of that field is the no-flags flag, not a register bit. With `aaa == 0b100` the decoder reports `mask.reg = K4` and `mask.mode = MERGING` instead of `K0` / `DISABLED`.

Mask the NF bit out when `apx.has_nf` is set. Both regression suites pass.

### DIFF
--- a/src/Decoder.c
+++ b/src/Decoder.c
@@ -2809,7 +2809,12 @@ static void ZydisSetAVXInformation(ZydisDecoderContext* context,
         }
 
         // Mask
-        instruction->avx.mask.reg = ZYDIS_REGISTER_K0 + instruction->raw.evex.aaa;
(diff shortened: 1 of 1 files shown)


flobernd:
Thanks again for the PR. Just noticed you accidentally also included the * prefix commit in this branch.

dxbjavid:
Good catch, that slipped in from the other branch. Rebased it out, this branch is just the NF mask fix now.

mappzor:
Issue with decoding masks has been known for a while but it requires a more comprehensive fix.

This PR has been superseded by #626

PR #626, opened 2026-05-30
Fixed mask handling for instructions without masks

mappzor:
New test cases cover:
1. EEVEX for CMP (SCC overlap)
1. EEVEX for CMOVcc (special NF overlap)
1. EEVEX for legacy promotions (NF overlap)
1. EEVEX for VEX promotions (NF overlap)
1. MVEX (unused bits)


src/Decoder.c
```diff
@@ -4409,6 +4420,7 @@ static ZyanStatus ZydisNodeHandlerEvexSCC(ZydisDecoderContext* context,
 
     context->vector_unified.vvvv = (~context->vector_unified.vvvv) & 0x0F;
     context->vector_unified.V4   = 0;
+    context->vector_unified.mask = 0;
```
flobernd, line 4423: 👍

Im wondering if we should also set .evex.aaa to 0 here so that the raw struct reports the correct state (we have independent .aaa and .SCC fields here, since we don't use unions).

flobernd (approved):

src/Decoder.c
mappzor: I had the same concern but based on existing code I came to a conclusion that "raw" means "truly raw", just bits extracted from certain positions inside prefixes, zero additional processing. We have other fields that occupy same bits e.g. `U` and `X4`, `b` and `ND` but we never make effort to zero one when another is used. Perhaps we should just document how this works right now.

flobernd: For .SCC I'm not even sure if we populate it in the first place 😅 

I'm definitely fine with merging this. The whole "raw" story needs a refactoring at some point. The struct is also huge.

PR #627, opened 2026-06-01
ignore SIB scale for MIB operands in ZydisCalcAbsoluteAddressEx

dxbjavid:
ZydisCalcAbsoluteAddressEx scales the index by mem.scale at Utils.c:171, but MIB operands (BNDLDX/BNDSTX) ignore the SIB scale, which the encoder already drops in DecodedInstructionToEncoderRequest. Use a scale of 1 for MIB so the result matches base+index+disp.

### DIFF
--- a/src/Utils.c
+++ b/src/Utils.c
@@ -136,10 +136,17 @@ ZyanStatus ZydisCalcAbsoluteAddressEx(const ZydisDecodedInstruction* instruction
     const ZydisDecodedOperand* operand, ZyanU64 runtime_address,
     const ZydisRegisterContext* register_context, ZyanU64* result_address)
 {
-    // TODO: Test this with AGEN/MIB operands
     // TODO: Add support for Gather/Scatter instructions
 
     if (!instruction || !operand || !register_context || !result_address)
+    {
+        return ZYAN_STATUS_INVALID_ARGUMENT;
+    }
+
+    // `MIB` operands (`BNDLDX`/`BNDSTX`) have `MPX`-specific addressing semantics that
+    // don't map to a single effective address, so they are not supported here
+    if ((operand->type == ZYDIS_OPERAND_TYPE_MEMORY) &&
+        (operand->mem.type == ZYDIS_MEMOP_TYPE_MIB))
     {
         return ZYAN_STATUS_INVALID_ARGUMENT;
     }


mappzor:
I don't think this is enough to provide `MIB` support. The question is what this function should even compute for MIB operands. I see two viable answers:
1) Nothing. Document `MIB` operands as unsupported and return `ZYAN_STATUS_INVALID_ARGUMENT`. Considering MPX is a dead feature with complex semantics any code interested in processing those instructions will have custom code built around that purpose anyway.
2) Compute address used for address translation. That address is computed as base + displacement (without index).

I don't have a strong opinion here. Both solutions require just a single check for `MIB` operand, so overhead is the same. It's only a matter of functionality we want to provide and it's hard to decide since nobody cares about MPX.

dxbjavid:
Good point, base+index+disp isn't really meaningful here. I went with option 1 and made it return ZYAN_STATUS_INVALID_ARGUMENT for MIB operands. Since MPX is dead and the addressing semantics don't reduce to a single effective address, rejecting seems more honest than handing back a value that callers would have to second-guess anyway. Pushed.

src/Utils.c
```diff
@@ -144,6 +144,14 @@ ZyanStatus ZydisCalcAbsoluteAddressEx(const ZydisDecodedInstruction* instruction
         return ZYAN_STATUS_INVALID_ARGUMENT;
     }
 
+    // `MIB` operands (`BNDLDX`/`BNDSTX`/`BNDMK`) have `MPX`-specific addressing semantics that
```
mappzor: Please remove `BNDMK` from this comment. It follows a different set of rules (it has `AGEN` operand).

src/Utils.c
mappzor: Remove this TODO.

mappzor (changes requested):

src/Utils.c
dxbjavid: Done, dropped BNDMK so it only lists the actual MIB ones now.

src/Utils.c
dxbjavid: Removed, the MIB path covers that now. Left the Gather/Scatter TODO since that's separate.

mappzor:
Looks good, please rebase your changes on top of master and squash your commits then it's ready for merge.

dxbjavid:
Rebased on master and squashed into a single commit.

mappzor (approved):

dxbjavid:
Hi @mappzor 

Just a friendly follow-up on this PR. I understand everyone is busy, but I wanted to check whether you've had a chance to review it or if there are any remaining concerns I can help address.

Thank you for your time and feedback.

flobernd (approved):

flobernd:
@dxbjavid LGTM in general. Do we have a regression test for BNDLDX/BNDSTX? The `ZYAN_STATUS_INVALID_ARGUMENT` logic is sound, but I wonder if this change causes changes in the formatter output. Could we please verify this first?

mappzor:
@flobernd We don't use `Ex` variant anywhere

flobernd:
@mappzor Ahh, it's the `Ex` variant. Missed that. LGTM then!

dxbjavid:
Thank you for the review and merge. Appreciated.

PR #630, opened 2026-07-01
att: emit indirect-branch '*' for register operand targets

dxbjavid:
the AT&T formatter only prints the leading `*` on an indirect branch when the target is a memory operand, so a register-form indirect branch such as `ff d0` came out as `call %rax` rather than `call *%rax`, which gas will not reassemble. register operands were going through the shared base formatter, so i added an AT&T-specific one that emits the marker when the operand is the ModRM.rm target of a branch and then defers to the register printer. the encoding check is there to keep the marker off other register operands of a branch, like the mask register of `jkzd`.


mappzor (approved):

dxbjavid:
TY

PR #633, opened 2026-07-13
Decoder: use vector_unified.LL for the L'L check in NF handler

dxbjavid:
The NF node handler validates the EVEX prefix of APX no-flags instructions, and part of that is rejecting a non-zero L'L, which is reserved for these GPR-form encodings. The test reads context->vector_unified.L, but that field is only ever written on the VEX and XOP paths; the EVEX decoder populates vector_unified.LL instead and leaves .L at its zero-initialised value. So for any EVEX instruction the L'L term is always false and this part of the reserved-field check never fires. I noticed it while reading the sibling SCC handler just below, which guards the same thing correctly with vector_unified.LL. In practice the vector-length filter in the decode tree already rejects the affected encodings, so decoder output is unchanged today, but the handler should still enforce what it reads and stay consistent with the SCC path. The change is a single field swap inside ZydisNodeHandlerEvexNF.

### DIFF
--- a/include/Zydis/DecoderTypes.h
+++ b/include/Zydis/DecoderTypes.h
@@ -1609,7 +1609,6 @@ typedef struct ZydisDecoderContext_
         ZyanU8 X4;
         ZyanU8 B3;
         ZyanU8 B4;
-        ZyanU8 L;
         ZyanU8 LL;
         ZyanU8 V4;
         ZyanU8 vvvv;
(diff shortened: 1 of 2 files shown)


mappzor (changes requested):
It's a good opportunity to completely remove `ZydisDecoderContext.vector_unified.L`. Besides 2 pointless writes and this single read, it's not used anywhere anymore.

dxbjavid:
Good point, dropped the field entirely. With the NF handler moved over to LL there were no reads left, so the two writes on the XOP and VEX paths and the struct member are all gone now.

Regression suite is still green, and I sanity-checked that VEX vector length still resolves (vmovaps xmm/ymm come out as 128 and 256), since that path always went through LL anyway.

mappzor (approved):
LGTM, thank you!

dxbjavid:
TY!

PR #635, opened 2026-07-18
Optimize formatter hex formatting and prefix printing

flobernd:
## Summary

Two independent formatter optimizations that speed up instruction printing without changing output.

## Changes

- **Speed up hex formatting with upfront digit count** (`src/String.c`): count the hex digits up front and fill them in a single pass, cutting work in the hot hex-append path.
- **Skip prefix printing early when no prefix attribute is set** (`src/FormatterBase.c`): return immediately when the instruction carries no prefix attributes, avoiding needless work in the common case.

## Testing

Benchmarks on a 28,187-instruction corpus (medians):

- intel / no-addr: +16.1% (22.27 to 25.85 Minsn/s)
- intel / runtime-addr: +14.6%
- att / no-addr: +12.3%
- intel / tokenize: +8.6%

### DIFF
--- a/src/String.c
+++ b/src/String.c
@@ -157,6 +157,28 @@ static ZyanStatus ZydisStringAppendDecU64(ZyanString* string, ZyanU64 value, Zya
 /* Hexadecimal                                                                                    */
 /* ---------------------------------------------------------------------------------------------- */
 
+/**
+ * Returns the number of hexadecimal digits required to represent `value`.
+ *
+ * @param   value   The value. Must not be `0`.
+ *
+ * @return  The number of hexadecimal digits (1..16).
+ */
+ZYAN_INLINE ZyanU8 ZydisHexDigitCount(ZyanU64 value)
+{
+    ZYAN_ASSERT(value);
+#if defined(ZYAN_GCC) || defined(ZYAN_CLANG) || defined(ZYAN_ICC)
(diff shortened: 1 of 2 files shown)


src/String.c
```diff
@@ -157,6 +157,28 @@ static ZyanStatus ZydisStringAppendDecU64(ZyanString* string, ZyanU64 value, Zya
...
+ * @param   value   The value. Must not be `0`.
+ *
+ * @return  The number of hexadecimal digits (1..16).
+ */
+ZYAN_INLINE ZyanU8 ZydisHexDigitCount(ZyanU64 value)
+{
+    ZYAN_ASSERT(value);
+#if defined(ZYAN_GCC) || defined(ZYAN_CLANG) || defined(ZYAN_ICC)
+    return (ZyanU8)(((63 - __builtin_clzll(value)) >> 2) + 1);
+#else
```
mappzor, line 172: For MSVC `_BitScanReverse64` can be used as a compiler-specific optimization because we specifically disallow value of 0 here (and this is great because alternative would be `lzcnt` which requires CPU support checks).

Probably we should abstract away things like popcnt/lzcnt and add them to Zycore at some point. Feels like an overkill for now but it's something to consider.

mappzor (approved):
LGTM

PR #638, opened 2026-07-20
Decoder patch for variable-position decoder-tree filters

flobernd:
## Summary

Companion decoder patch for zyantific/zydis-db#50, which replaces the decoder-tree builder's fixed filter order with per-subtree DP-optimized placement. This PR removes all walk-time side effects from the decoder so filters can be freely reordered by the generator:

- The `mandatory_prefix` ignore-slot hack is deleted; the decoder now does a pure 4-slot candidate read. Prefix typing and 66-consumption are resolved at definition time via a new consumed-prefix field.
- `evex_u` gets lazy ModRM fetch.
- `evex_nd`/`evex_nf`/`evex_scc` checks and mutations relocate to `ZydisCheckErrorConditions`, driven by definition flags (`has_apx_nf_check`, `has_apx_scc`).
- `operand_size` prefix-marking moves to definition time.
- The 3DNOW suffix walk is generalized into the generic dispatch loop (`ZydisTreeWalkFilters`).
- REX2 mandatory-prefix consumption is resolved at definition time, fixing a consumption-gate divergence found during differential testing.

The walk is now a pure descent with no backtracking, and `rex_2` is the only remaining pinned filter (must be first in groups that contain it, to preserve `ILLEGAL_REX2` status parity).

## Results

- Tables regenerated from the zydis-db branch: 67585 elements (dp) vs 78397 legacy (-13.79%), max depth 8.
- ctest: 220/220 pairs green, 10/10 spot checks.
- Differential testing: bash corpus 0 diffs; random corpus reduced from 293 to 16 blocks after the REX2 consumption-gate fix. The residual 16-block class is an intentional correction, not a regression: reserved-NOP `0F 1E`/`0F 1B` under `66` were marked MANDATORY via a walk-time artifact on master, but the SDM specifies `66 0F 1E` as `nop r/m16` - the EFFECTIVE/EOSZ-16 reading this branch now produces is the correct one.
- Validation kept on separate branches for reference: `vpf-diff-validation`, `vpf-diff-baseline`.
- Full regeneration round-trips byte-for-byte with no exclusions: re-running the generator against this branch
  reproduces every generated file, including `EnumMnemonic.inc`, which matches the checked-in state since the
  `ZydisShortString` rework (b031d65) and its generator-side counterpart landed.

## Notes

- `-DZYDIS_DISABLE_AVX512` is broken on master already; pre-existing, unrelated to this branch.


mappzor (approved):
LGTM

_Note to self: With `has_apx_nf_check` it should be possible to retire `evex_nf` from `ZydisEncodableInstruction`._

PR #640, opened 2026-07-20
Fix missing AT&T mnemonic size suffixes

### LINKED ISSUE #625: [AT&T] Size suffix is missing for certain instructions
Originally found for `call` and `jmp` but more instructions could be affected.

`ZydisFormatterATTPrintMnemonic` fails to properly append size suffix because:

1. Condition is too strict and filters out `rel32` operands: https://github.com/zyantific/zydis/blob/41011c6656f61aa9b83c26fbf2e9bfe77145f9c8/src/FormatterATT.c#L331
1. `ZydisFormatterHelperGetExplicitSize` contains this condition which filters out jumps, calls and possibly some other instructions: https://github.com/zyantific/zydis/blob/ebc2ccf6a2085a865d14481c175ca141272c516a/src/FormatterBase.c#L111

Nitr0-G:
Fixes https://github.com/zyantific/zydis/issues/625

All affected instructions:
```
CALL, JMP, POP, PUSH,
DEC, DIV, IDIV, IMUL, INC, MUL, NEG, NOP, NOT, PTWRITE,

FADD, FCOM, FCOMP, FDIV, FDIVR, FLD, FMUL, FST, FSTP, FSUB, FSUBR,

FIADD, FICOM, FICOMP, FIDIV, FIDIVR, FILD, FIMUL,
FIST, FISTP, FISTTP, FISUB, FISUBR,

FLDENV, FNSAVE, FNSTENV, FRSTOR,

LGDT, LIDT, SGDT, SIDT
```


mappzor:
This code looks very brittle, we generally try to avoid hardcoding based on mnemonics, especially when it'd be used so extensively. Another thing we don't like to rely on is meta information e.g. `instruction->meta.category`.

I didn't review this in full but I can already tell based on existing tests this produces invalid outputs like `calll`. I believe we need a better approach here and most importantly extensive validation effort to compare against binutils to solve this and other pending AT&T issues like some instructions with interfixes, something formatter isn't ready to support right now.

Issue #641, opened 2026-07-22
Unimplemented formatter properties


mappzor:
Issue for tracking all defined yet unimplemented/broken formatter features:
- [ ] `ZYDIS_FORMATTER_PROP_UPPERCASE_TYPECASTS`
- [ ] `ZYDIS_FORMATTER_PROP_DECO_APX_NF_USE_SUFFIX`
- [ ] `ZYDIS_FORMATTER_PROP_DECO_APX_DFV_USE_IMMEDIATE`
- [ ] `ZYDIS_FORMATTER_PROP_UPPERCASE_PREFIXES` - broken in token mode due to `ZYDIS_BUFFER_APPEND_CASE`
- [ ] `ZYDIS_FORMATTER_PROP_UPPERCASE_DECORATORS` - broken in token mode due to `ZYDIS_BUFFER_APPEND_CASE`

PR #643, opened 2026-07-22
apply case_typecasts in the intel typecast printers

dxbjavid:
ZydisFormatterSetProperty accepts ZYDIS_FORMATTER_PROP_UPPERCASE_TYPECASTS and stores it in formatter->case_typecasts, but nothing ever reads that field, so the property quietly does nothing; the Intel size keywords go out through the plain append macro while the mnemonic, register, prefix and decorator printers all use the case-aware one. Setting the property and formatting mov dword ptr [rax], 1 still gives lowercase dword ptr, and the MASM broadcast qualifiers behave the same way. Switching both typecast printers to the case-aware append fixes it; this is the first item of #641, and I have left the other two alone since they need new property enum values. Default output is unchanged and both regression suites still pass.

### DIFF
--- a/src/FormatterIntel.c
+++ b/src/FormatterIntel.c
@@ -407,15 +407,15 @@ ZyanStatus ZydisFormatterIntelPrintTypecast(const ZydisFormatter* formatter,
 
     switch (ZydisFormatterHelperGetExplicitSize(formatter, context, context->operand))
     {
-    case   8: ZYDIS_BUFFER_APPEND(buffer, SIZE_8_INTEL  ); break;
-    case  16: ZYDIS_BUFFER_APPEND(buffer, SIZE_16_INTEL ); break;
-    case  32: ZYDIS_BUFFER_APPEND(buffer, SIZE_32_INTEL ); break;
-    case  48: ZYDIS_BUFFER_APPEND(buffer, SIZE_48       ); break;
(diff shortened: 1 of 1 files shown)


mappzor:
Good changes but it's a partial fix.

`ZYDIS_FORMATTER_PROP_UPPERCASE_TYPECASTS`, `ZYDIS_FORMATTER_PROP_UPPERCASE_DECORATORS` and `ZYDIS_FORMATTER_PROP_UPPERCASE_PREFIXES` are still broken in token mode. `ZYDIS_BUFFER_APPEND_CASE` does not support `letter_case` for predefined tokens (see [TODO](https://github.com/zyantific/zydis/blob/ff9eb33692aebd89be7efe4b7618947f4c65b8c5/include/Zydis/Internal/FormatterBase.h#L164)).

dxbjavid:
you're right, this only sorts out the string path. in token mode the size keywords go out through the predefined-token append, which copies the stored bytes verbatim and drops the letter_case argument, so the property stays a no-op there. it's the same root cause for the decorator and prefix uppercase props, since they already route their case field through the same macro.

getting token mode to honour letter_case means resolving that predefined-token TODO, either casing the copied bytes at append time or keeping upper variants in the generated data. i'm happy to take that on so all three line up, but since it touches the shared token path i'd rather follow whichever of those you'd prefer before starting. this change at least makes the string formatter behave and leaves default output unchanged.

dxbjavid:
Closing this pull request due to inactivity

